PR #1, opened 2022-06-22
smolyak index grid

mzf0069:
Added a method to generate the unique indexes for each level. 
Added a method to generate the index of grid points.


cakieslich:
Hello world

smolyay/smolyak.py
```diff
@@ -0,0 +1,158 @@
...
+        self.index_per_level = index_per_level
+
+        if (type(index_per_level) == tuple or
+                type(index_per_level) == np.ndarray):
+            index_per_level = list(index_per_level)
+
+        if len(index_per_level) != exactness + 1:
+            raise IndexError(
+                "index_per_level must be an array with a length of {}"
+                .format(exactness + 1))
```
mzbush, line 30: Technically, the index_per_level can describe more levels than is required for the exactness. If you only want to generate the index_per_level once and use that for multiple values of exactness, you can just generate it for the highest value for exactness and then never need to generate it again. This should only check if the index_per_level describes less than the required amount.

If the index_per_level cannot contain extra information, then a check could be used to take only the required portion in a new list.

smolyay/smolyak.py
```diff
@@ -0,0 +1,158 @@
...
+        Returns
+        -------
+        level_indexes : numpy array
+            Unique indexes at each level.
+
+        """
+        # creating an array to store unique indexes in each level
+        level_indexes = np.zeros((self.exactness+1), dtype=list)
+
+        max_index = sum(self.index_per_level)
```
mzbush, line 44: Here the code requires that index_per_level cannot contain extra information. This part might be edited to avoid this.

smolyay/smolyak.py
```diff
@@ -0,0 +1,158 @@
...
+            level_indexes[idx_index_per_level] = (
+                indexes[summ_index-self.index_per_level[idx_index_per_level]:
+                        summ_index])
+
+        return level_indexes
+
+    def grid_point_index(self):
+        """Generate index of grid points.
+
+        First, it generates allowed composoitons of levels
```
mzbush, line 63:
```suggestion
        First, it generates allowed compositions of levels
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,158 @@
...
+        summs = np.linspace(self.dimension,
+                            self.exactness+self.dimension,
+                            self.exactness+1, dtype=np.int32)
+        max_level_indexes = summs - self.dimension
+        num_comps = 0
+        for max_level_index in max_level_indexes:
+            num_comps += scipy.special.comb(max_level_index+self.dimension-1,
+                                            self.dimension-1, exact=True)
+        level_compositions = np.zeros((num_comps, self.dimension), dtype=list)
+
```
mzbush, line 87: For the possible level_compositions, is there ever going to be a reason that we want the code to print that out so it can be seen for writing a paper or any other reason? If so, this function may need to be split into multiple functions.

smolyay/smolyak.py
```diff
@@ -0,0 +1,158 @@
...
+            # corresponding to each level
+            for level_index in level_index_composition:
+                len_level.append(len(level_index))
+            # generate grid point indexes
+            for index_grid_i in np.ndindex(tuple(len_level)):
+                for dimension_i in range(self.dimension):
+                    index = index_grid_i[dimension_i]
+                    grid_points_index[num_grid_point][dimension_i] = (
+                        level_index_composition[dimension_i][index])
+                num_grid_point += 1
```
mzbush, line 155: This for loop is very nested. May want to avoid that for optimization purposes.

mzbush (changes requested):
Most things look good, besides a few tweaks to make it better.

smolyay/smolyak.py
mzf0069: Addressed.

smolyay/smolyak.py
mzf0069: Addressed.

smolyay/smolyak.py
mzf0069: For the case of generating grid points, I don't think it's necessary; however, we can split it in case it's needed.

mzbush (approved):
Everything looks fine. I am happy.

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+    def level_indexes(self):
+        """Generate indexes of levels.
+
+        Returns
+        -------
+        level_indexes : list
+            Unique indexes at each level.
+
+        """
+        # creating an array to store unique indexes in each level
```
mphoward, line 38: I think this whole function can all be done in a more readable way like this
```py
# cumulative sum to get end index of each level, up to exactness+1
end_levels = numpy.cumsum(self.index_per_level[:self.exactness+1])
# create ranges of indexes at each level
level_indexes = [list(range(end-n, end)) for end, n in zip(end_levels, self.index_per_level)]
return level_indexes
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+        summ_index = self.index_per_level[0]
+        for idx_index_per_level in range(1, self.exactness + 1):
+            summ_index += self.index_per_level[idx_index_per_level]
+            level_indexes[idx_index_per_level] = (
+                indexes[summ_index-self.index_per_level[idx_index_per_level]:
+                        summ_index])
+
+        return level_indexes
+
+    def grid_point_index(self):
```
mphoward, line 56: This will also become a property, so it should get a short docstring like above. The comments about NEXCOM, etc. belong up in the detailed class description, which still needs to be written.

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+
+        """
+        num_grid_points = 0
+        level_indexes = self.level_indexes()
+        level_compositions = []
+
+        # compute the allowed level compositions
+        for summ in range(self.dimension, self.exactness+self.dimension+1):
+
+            # NEXCOM
```
mphoward, line 77: As we discussed, we should probably pull this out into a function

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
+import numpy as np
```
mphoward, line 1: In published code, you should try not to alias packages unless you really need to because it pollutes the namespace:
```suggestion
import numpy
```
In your own scripts, it's perfectly fine to do this though

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+                t = r[h-1]
+                r[h-1] = 0
+                r[0] = t-1
+                r[h] += 1
+
+                # (D)
+                compositions[index] = r
+                index += 1
+
+            # precompute the number of grid points
```
mphoward, line 111: I have a really hard time following the code from here onward, but I think that what you are doing can be achieved in a simpler way using `numpy.meshgrid`. See https://stackoverflow.com/questions/1208118/using-numpy-to-build-an-array-of-all-combinations-of-two-arrays for an explanation:

You would define `points = None` at the start of the function, then do something like this inside the for loop:
```py
            # get all combinations of points at each level
            for composition in compositions:
                point_ids = [level_indexes[i] for i in composition]
                points_ = np.array(np.meshgrid(*point_ids)).T.reshape(-1, self.dimension)
                if points is None:
                    points = points_
                else:
                    points = np.concatenate((points,points_), axis=0)
```
`points` will be a NumPy array, so you can either convert to a list using the `tolist()` method or leave as is

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
+import numpy as np
+import scipy.special
+
+
+class IndexGrid:
+    """Generate index of grid points.
```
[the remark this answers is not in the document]
mzf0069: Done.

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, dimension, exactness, index_per_level):
+        self.dimension = dimension
+        self.exactness = exactness
+        self.index_per_level = index_per_level
+
+        if len(index_per_level) < exactness + 1:
+            raise IndexError(
+                "index_per_level must be an array with a"
+                "length of at least {}"
+                .format(exactness + 1))
```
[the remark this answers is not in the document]
mzf0069: Created a `_update` function and moved it there.

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def level_indexes(self):
+        """Generate indexes of levels.
+
+        Returns
+        -------
+        level_indexes : list
+            Unique indexes at each level.
+
+        """
```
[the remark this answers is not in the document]
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
```diff
@@ -0,0 +1,141 @@
+import numpy as np
+import scipy.special
+
+
+class IndexGrid:
+    """Generate index of grid points.
+
+    Attributes
```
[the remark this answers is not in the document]
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+        r[h] += 1
+
+        # (D)
+        compositions[index] = r
+        index += 1
+
+    return compositions
+u = IndexGrid(2, 2, [1,2,2])
+i = u.level_indexes
+j = u.grid_point_index
```
mphoward, line 157: Don't commit debugging, these belong in tests though!
```suggestion
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+            Index of grid points
+
+        """
+        if self._needs_update:
+            self._update()
+
+        return self._grid_point_index
+
+
+def generate_compositions(summ, dimension):
```
mphoward, line 106: Function arguments should be descriptive and make sense for the function. In this case `generate_compositions` is a function that just does NEXCOM, and it is not really tied to the `IndexGrid` class. You could use `n, k` to match the variables used in the algorithm or `value, num_parts` if you want to be more descriptive.

Further, I think you should add a third boolean argument like `weak` or `include_zero`. If `True`, this function should include zeros in the composition.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+    r = numpy.zeros(dimension, dtype=numpy.int32)
+    r[0] = max_level_index
+    t = max_level_index
+    h = 0
+    compositions[0] = r
+    index = 1
+
+    # (D): these termination conditions should be redundant
+    while (r[dimension-1] != max_level_index
+           and index < num_output + 1):
```
mphoward, line 137: I think the `+1` is incorrect here. I would also only use one check or the other, as these are redundant. Technically, it would be faster to use only the first one, since then you don't have to precompute the number of compositions.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # (C)
+        h += 1
+        t = r[h-1]
+        r[h-1] = 0
+        r[0] = t-1
+        r[h] += 1
+
+        # (D)
+        compositions[index] = r
```
mphoward, line 151: Rather than save this into an array, you can use Python's `yield` syntax to turn this into a generator. That way, you don't actually have to generate all the compositions in memory (if you don't want to).
```suggestion
        yield r
```
You would then delete the return statement below and the `compositions` variable above. This is convenient because you can easily do:
```
for c in generate_compositions(n,k):
    print c
```
without generating all the compositions at once, but also
```
compositions = list(generate_compositions(n,k))
```
if you do want them.

Don't forget to add 1 to everything here if you are doing strong composition (no zeros).

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+    Returns
+    -------
+    compositions: list
+        Allowed composition of levels.
+    """
+    max_level_index = summ - dimension
+    num_output = scipy.special.comb(max_level_index+dimension-1,
+                                    dimension-1, exact=True)
+    compositions = numpy.zeros((num_output, dimension),
+                               dtype=numpy.int32)
```
mphoward, line 125: These can be removed with the suggestions below

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # (A) first entry
+    r = numpy.zeros(dimension, dtype=numpy.int32)
+    r[0] = max_level_index
```
mphoward, line 129: I would suggest to use variable names consistent with the paper the algorithm came from, it is easier to read later on.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+        Summation of levels.
+    dimension: int
+        Number of independent variables.
+
+    Returns
+    -------
+    compositions: list
+        Allowed composition of levels.
+    """
+    max_level_index = summ - dimension
```
mphoward, line 121: Here, you would use the third (boolean) argument to decide what the max value used by the algorithm was

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # cumulative sum to get end index of each level, up to exactness+1
+        end_levels = numpy.cumsum(self.index_per_level[:self.exactness+1])
+        # create ranges of indexes at each level
+        level_indexes = [list(range(end-n, end))
+                         for end, n in zip(end_levels, self.index_per_level)]
+        self._level_indexes = level_indexes
+
+        points = None
+        # get all combinations of points at each level
```
mphoward, line 55: Stylistically, I like the code under a comment to all appear together
```suggestion
        # get all combinations of points at each level
        points = None
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+        for summ in range(self.dimension, self.exactness+self.dimension+1):
```
mphoward, line 56: You should either use a more descriptive variable name than `summ` or a traditional index variable like `i`.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+            # use generate_compositions function which implements NEXCOM
```
mphoward, line 57: Per PEP 8 (I think), don't write comments for things that are obvious
```suggestion
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+            compositions = generate_compositions(summ, self.dimension)
+            for composition in compositions:
```
mphoward, line 59: If you use the generator style I suggested, this will work:
```suggestion
            for composition in generate_compositions(summ, self.dimension, include_zeros=False):
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+                point_ids = [level_indexes[i] for i in composition]
```
mphoward, line 60: Is `point_ids` the right name for this? I just chose something random, so you should check that. Also, be careful here because `i` is used as an iterator inside an outer loop. You might want to switch to `j` or something else like `c` to make sure it doesn't accidentally get used

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+                points_ = numpy.array(
```
mphoward, line 61: A comment would be helpful here that this is a code for generating all combinations of the arrays along each dimension

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+                if points is None:
+                    points = points_
+                else:
+                    points = numpy.concatenate((points, points_), axis=0)
+        self._grid_point_index = points
+        self._needs_update = False
+
+    @property
+    def level_indexes(self):
+        """Generate indexes of levels.
```
mphoward, line 73: The docstring needs to be shortened for both properties (see old suggestion). Properties generally only get one-liners.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
...
+    exactness : int
+        Level of exactness of the approximation.
+    index_per_level : list
+        Number of indexes at each level.
+    """
+
+    def __init__(self, dimension, exactness, index_per_level):
+        self.dimension = dimension
+        self.exactness = exactness
+        self.index_per_level = index_per_level
```
mphoward, line 29: You will want to turn each of these into a property as well, which has both a getter and a setter. The setters should make `self._needs_update = True`.

smolyay/smolyak.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Jul  5 08:00:15 2022
+
+@author: mzf0069
+"""
+
```
mphoward, line 8: Do not include these in your files. The top part is not necessary, and we will write docstrings for the modules.

mphoward (changes requested):
This is a good improvement! Here are some further comments and suggestions

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done. 

Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mzf0069: In the paper, they used simply "n" instead of "max_level_index". However, since "n" differs depending on zero being included or not, I named it max_value instead.

smolyay/smolyak.py
mzf0069: Done.

smolyay/smolyak.py
mphoward: Right, but `max_level_index` is meaningless to `generate_compositions` because this function doesn't know anything about `IndexGrid`. The math problem is usually defined as "the compositions of n into k parts", which is why NEXCOM uses the names it does. I suggest the same here.

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+    @dimension.setter
+    def dimension(self, value):
+        print('Set dimension')
+        self._dimension = value
+        self._needs_update = True
+
+    @property
+    def exactness(self):
+        """Get exactness."""
```
mphoward, line 35: Or this:
```suggestion
        """int: Level of exactness."""
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+        print('Set index_per_level')
+        self._index_per_level = value
+        self._needs_update = True
+
+    def _update(self):
+        """Update the indexes of grid points.
+
+        Generates the indexes for each level
+        depending on the index_per_level, and makes the
+        indexes of grid points.
```
mphoward, line 60: You should document any exceptions that you might raise (e.g., IndexError) here in a `Raises` section.

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+        end_levels = numpy.cumsum(self.index_per_level[:self.exactness+1])
+        # create ranges of indexes at each level
+        level_indexes = [list(range(end-n, end))
+                         for end, n in zip(end_levels, self.index_per_level)]
+        self._level_indexes = level_indexes
+
+        # get all combinations of points at each level
+        grid_points_idx = None
+        for sum_levels in range(self.dimension,
+                                self.exactness+self.dimension+1):
```
mphoward, line 79: This makes it clearer that the range is from dimension up:
```suggestion
                                self.dimension+self.exactness+1):
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+        # cumulative sum to get end index of each level, up to exactness+1
+        end_levels = numpy.cumsum(self.index_per_level[:self.exactness+1])
+        # create ranges of indexes at each level
+        level_indexes = [list(range(end-n, end))
+                         for end, n in zip(end_levels, self.index_per_level)]
+        self._level_indexes = level_indexes
+
+        # get all combinations of points at each level
+        grid_points_idx = None
+        for sum_levels in range(self.dimension,
```
mphoward, line 78: An iterator should be singular
```suggestion
        for sum_level in range(self.dimension,
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+    def grid_point_index(self):
+        """Generate index of grid points."""
+        if self._needs_update:
+            self._update()
+
+        return self._grid_point_index
+
+
+def generate_compositions(value, num_parts, include_zero):
+    """Genearte compostions of a value into num_parts parts.
```
mphoward, line 116:
```suggestion
    """Generate compositions of a value into parts.
```

smolyay/smolyak.py
```diff
@@ -0,0 +1,156 @@
...
+        All possible compositions of the value into num_parts.
+    """
+    value = value if include_zero else value - num_parts
+
+    # (A) first entry
+    r = numpy.zeros(num_parts, dtype=numpy.int32)
+    r[0] = value
+    t = value
+    h = 0
+    index = 1
```
mphoward, line 139: I think this variable is not needed?
```suggestion
```

mphoward (changes requested):
Great work! Here are a few more small things, but then this mostly just needs testing.

tests/smolyay_test.py
```diff
@@ -0,0 +1,200 @@
+import pytest
+import numpy
+from smolyay.smolyak import IndexGrid
+
+
+grid_points_expect = numpy.array([[0,0],[0,1],[0,2],[0,3],[0,4],[1,0],[1,1],[1,2],[2,0],[2,1],[2,2],[3,0],[4,0]], dtype=numpy.int32)
```
mphoward, line 6: Rather than make this a global variable, you can make it a test fixture:
```py
@pytest.fixture
def expected_points():
    return numpy.array(...)

def test_something(expected_points):
    assert point in expected_points
```

tests/smolyay_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+def test_grid_point_index():
+    """test grid point index"""
+    test_class = IndexGrid(2,2,[1,2,2])
+    grid_points = test_class.grid_point_index
+    # sort grid_points as order of points is unimportant
+    for i in range(len(grid_points[0,:])-1,-1,-1):
+        grid_points = grid_points[grid_points[:,i].argsort(kind='mergesort')]
+
+    assert numpy.array_equal(grid_points,grid_points_expect)
```
mphoward, line 55: Can't this be done with:
```
    assert sorted(test_class.grid_point_index) == grid_points_expect
```
?

tests/smolyay_test.py
```diff
@@ -0,0 +1,200 @@
...
+    #include ways test failed if it did
+    error_message = []
+    if lpi_12:
+        error_message = "New index per level failed to update value. "
+    elif not lpi_2e:
+        error_message = "New index per level has incorrect value. "
+
+    assert not error_message,error_message
+
+def test_error():
```
mphoward, line 196: Give this a more descriptive name: what is it testing for an error in?

mphoward (changes requested):
Thank you! I have some comments on both the main code and the tests. It would be good if you both can make a plan to implement these.

mphoward:
I will wait for @mzbush to update the tests before I review, as I think that should be close to final pass. Please let me know when you're ready for that!

tests/smolyay_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+def test_initial_exactness():
+    """test initialized class returns correct exactness"""
+    test_class = IndexGrid(1,2,[1,2,3])
+    assert test_class.exactness == 2
+
+def test_initial_index_by_level():
+    """test initialized class returns correct index by level"""
+    test_class = IndexGrid(1,2,[1,2,3])
+    assert numpy.array_equiv(test_class.index_per_level,[1,2,3])
```
mphoward, line 29: I think that `array_equiv` is a weaker test so this should be `array_equal`. Was it failing otherwise?
```suggestion
    assert numpy.array_equal(test_class.index_per_level,[1,2,3])
```
In general, these tests make clear that there is a mixture of lists and NumPy arrays being used in the class. Do we want to perhaps use only lists? That would allow these tests to simplify to using `==` operators, and I could imagine that being helpful elsewhere. It should be as easy as converting the NumPy array `tolist` before storing it in the class.

tests/smolyay_test.py
```diff
@@ -0,0 +1,159 @@
...
+    assert numpy.array_equal(grid_point_1,grid_point_2)
+
+def test_increase_exactness(expected_points):
+    """test if grid points are correct if exactness is increased"""
+    test_class = IndexGrid(2,1,[1,2,2])
+    grid_points_1 = sorted(test_class.grid_point_indexes,key = lambda x:x[0])
+    test_class.exactness = 2
+    grid_points_2 = sorted(test_class.grid_point_indexes,key = lambda x:x[0])
+    assert ((not numpy.array_equiv(grid_points_1,grid_points_2))
+            and numpy.array_equiv(grid_points_2, expected_points))
```
mphoward, line 76: Perform each of these assertions separately
```suggestion
    assert not numpy.array_equiv(grid_points_1,grid_points_2)
    assert numpy.array_equiv(grid_points_2, expected_points)
```

tests/smolyay_test.py
```diff
@@ -0,0 +1,159 @@
...
+        [2,4],[1,5]],dtype=numpy.int32)
+    composition_obtained = list(generate_compositions(6,2,include_zero=False))
+    assert numpy.array_equiv(composition_obtained,composition_expected) 
+
+def test_generate_compositions_zero_false_error():
+    """test that generate compositions raises an error for invalid input"""
+    with pytest.raises(ValueError):
+        composition_obtained = generate_compositions(6,7,include_zero=False)
+        for obj in composition_obtained:
+            pass
```
mphoward, line 159: This would be an easier way to run the generator:
```suggestion
        list(generate_compositions(6,7,include_zero=False))
```

mphoward (changes requested):
A couple more small changes from my side, then I think this is ready for Dr. K to review and merge

tests/smolyay_test.py
```diff
@@ -0,0 +1,156 @@
...
+def test_initial_index_by_level():
+    """test initialized class returns correct index by level"""
+    test_class = IndexGrid(1,2,[1,2,3])
+    assert test_class.index_per_level == [1,2,3]
+
+def test_level_indexes(expected_indexes):
+    """test level indexes"""
+    test_class = IndexGrid(2,2,[1,2,2])
+    level_indexes_recieved = test_class.level_indexes
+
```
mphoward, line 34:
```suggestion
```

tests/smolyay_test.py
```diff
@@ -0,0 +1,156 @@
...
+    test_class = IndexGrid(2,2,[1,2,2])
+    level_indexes_recieved = test_class.level_indexes
+    
+    assert level_indexes_recieved == expected_indexes
+
+def test_grid_point_index(expected_points):
+    """test grid point index"""
+    test_class = IndexGrid(2,2,[1,2,2])
+    grid_points = sorted(test_class.grid_point_indexes,key = lambda x:x[0])
+
```
mphoward, line 41:
```suggestion
```

tests/smolyay_test.py
```diff
@@ -0,0 +1,156 @@
...
+    assert grid_points == expected_points
+
+def test_level_indexes_extra_level_per_index(expected_indexes):
+    """test level indexes when level_per_index has more information"""
+    test_class = IndexGrid(2,2,[1,2,2,4,8])
+    level_indexes_recieved = test_class.level_indexes
+
```
mphoward, line 48:
```suggestion
```

tests/smolyay_test.py
```diff
@@ -0,0 +1,156 @@
...
+    
+    assert grid_point_1 == grid_point_2
+
+def test_increase_exactness(expected_points):
+    """test if grid points are correct if exactness is increased"""
+    test_class = IndexGrid(2,1,[1,2,2])
+    grid_points_1 = sorted(test_class.grid_point_indexes,key = lambda x:x[0])
+    test_class.exactness = 2
+    grid_points_2 = sorted(test_class.grid_point_indexes,key = lambda x:x[0])
+    assert not grid_points_1 == grid_points_2
```
mphoward, line 74:
```suggestion
    assert grid_points_1 != grid_points_2
```

mphoward (approved):
Great job! I have marked a couple minor final changes, which you can batch apply and commit using the GitHub web interface if you agree. Otherwise, lgtm.

@cakieslich when you review and approve, I recommend squash-merging this given the long commit history. @mzf0069 and @mzbush , if this is how this branch is merged, we may need to do a rebase of your other PRs before they can be merged. We can discuss how to do that later.

tests/smolyay_test.py
```diff
@@ -0,0 +1,156 @@
...
+    """test the level indexes if exactness is zero"""
+    test_class = IndexGrid(2,0,[1,2,2])
+    assert test_class.grid_point_indexes == [[0,0]]
+
+def test_generate_compositions_include_zero_true():
+    """test the generate compositions function if include_zero is true"""
+    composition_expected = [[6,0],[5,1],[4,2],[3,3],[2,4],[1,5],[0,6]]
+    composition_obtained = []
+    for i in generate_compositions(6,2,include_zero=True):
+        composition_obtained.append(i.copy())
```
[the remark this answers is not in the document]
mzbush: Funny thing about this part of the code. When I was making it for some reason using the list method didn't work when include_zero was true but it did work when include_zero was false. @mzf0069 and I have no idea why this is the case. That's why composition_obtained is generated differently.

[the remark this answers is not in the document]
mphoward: Ah, that indicates that there is an issue with the code to resolve. If the test is valid (and the change I suggested is certainly the way a generator is expected to work), you should modify the code to pass the test, not hack the test until the code works. Let us know if you need help debugging.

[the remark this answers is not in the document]
mzbush: Should I update the test or wait for @mzf0069 to fix the code so it passes the test?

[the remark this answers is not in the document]
mphoward: It doesn’t matter to me! One of you could take care of both, you do not need to be separate.

mphoward:
I believe this PR is ready to squash merge @cakieslich

cakieslich (approved):

PR #2, opened 2022-07-14
Feature/basis function

mzbush:
The BasisFunction class provides a framework for how the basis functions used to build the surrogate function should be treated. The first child class of BasisFunction is ChebyshevFirstKind, which describes the Chebyshev polynomials of the first kind.


basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
+from abc import ABC, abstractmethod
```
mphoward, line 1: For modules with short names, prefer to just import to avoid polluting the namespace:
```suggestion
import abc
```

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+
+class BasisFunction(ABC):
+    """Abstract class for basis functions.
+
+    Creates a framework for basis functions that are the building blocks
+    for the terms in the surrogate model. 
+    ''max_exactness'' represents the maximum level of exactness, or
```
mphoward, line 8: These should be backticks (not quotes)
```suggestion
    ``max_exactness`` represents the maximum level of exactness, or 
```

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Parameters
+    ----------
+    max_exactness : int
+        level of exactness class will describe
+
+    """
+
+    def __init__(self,max_exactness):
+        """Initializer"""
```
mphoward, line 31: You don't need a docstring for the `__init__` method if you are documenting the class as well.

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+        self._extrema_list = [0]
+        self._extrema_per_level_num = [1]
```
mphoward, line 33: This assumption that the first extrema is zero seems very restricting to me. I would like to see the code refactored in a way where the deriving classes get full control over where their extrema are located.

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+        self._max_exactness = max_exactness
+        self.update_extrema_index_per_level()
+
+    @property
+    def max_exactness(self):
+        """Maximum level of exactness described"""
+        return self._max_exactness
+
+    @property
+    def extrema_list(self):
```
mphoward, line 43: Property names should not include the data type, that is an implementation detail:
```suggestion
    def extrema(self):
```

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+        """Extrema assigned to Smolyak indices"""
+        return self._extrema_list
+
+    @property
+    def extrema_per_level_num(self):
```
mphoward, line 48:
```suggestion
    def num_extrema_per_level(self):
```
It looks like you are still missing the `extrema_per_level` variable (or whatever you want to call it) we talked about to identify which indices belong at each level? I still think this is more useful than just having the number at each level, as the number can be generated from that variable.

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+        Sets maximmum level of exactness class instance will describe, will
+        extend or truncate extrema_list and extrema_per_level_num to match
+        new value
+
+        Parameters
+        ----------
+        max_exactness : int
+            new level of exactness class should describe
+        """
+        if self._max_exactness < max_exactness:
```
mphoward, line 64: I would flip these so that it's clear what role the parameter plays in the function
```suggestion
        if max_exactness > self._max_exactness:
```

basis/BasisFunction.py
```diff
@@ -0,0 +1,110 @@
...
+    def update_extrema_index_per_level(self):
+        """Updates properties described by level of exactness
+        Compute extrema of basis function and the number of extrema
+        associated with each level of exactness required to describe 
+        the level of exactness in property self.max_exactness
+        """
+        pass
+
+    @abstractmethod
+    def basis_fun(x,n):
```
mphoward, line 89: You should not use abbreviations like `fun`. How about this instead? 
```suggestion
    def basis(x,n):
```

basis/ChebyshevFirstKind.py
```diff
@@ -0,0 +1,62 @@
...
+    following recurrence relation:
+    .. math::
+    T_0(x) = 1
+    T_1(x) = x
+    T_{n+1}(x) = 2xT_n(x) - T_{n-1}(x)
+
+    The extrema of the polynomials are calculated via the following
+    equation:
+    .. math::
+    extrema = -\cos\left((\frac{j-1}{n-1}\pi\right), j = 1,...,n
```
mphoward, line 17: You should choose a symbol for the j-th extrema at the n-th level, like:
```
x_{n,j}^*
```

basis/ChebyshevFirstKind.py
```diff
@@ -0,0 +1,62 @@
...
+
+    """
+
+    def update_extrema_index_per_level(self):
```
mphoward, line 21: This is where you would want to fix the issue of the first extrema being zero.

basis/ChebyshevFirstKind.py
```diff
@@ -0,0 +1,62 @@
...
+        -------
+        y : float
+            output
+        """
+        if n == 0:
+            return 1
+        elif n == 1:
+            return x
+        else:
+            k_lim = math.floor(n/2)
```
mphoward, line 58: Integer division might be better here, I forget if `math.floor` returns a float or int.
```suggestion
            k_lim = n//2
```

mphoward (changes requested):
Good start Michaela! Here are some suggestions.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
+import abc
```
mphoward, line 1: This file should be in `smolyay/basis.py` so that it can be imported like:
```py
import smolyay.basis
```
You should `git mv` this file, then `git rm` the `basis/` directory and `__init__.py`.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+import numpy
+import math
```
mphoward, line 3: Per [PEP-8](https://peps.python.org/pep-0008/#imports), imports should come in the order: standard library, third-party, local with a blank line between groups.
```suggestion
import math

import numpy
```

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+    grid level.
+
+    Parameters
+    ----------
+    max_exactness : int
+        level of exactness class will describe
+
+    """
+
+    def __init__(self,max_exactness):
```
mphoward, line 29: After our meeting last week, I don't think the base class `BasisFunction` should have `max_exactness` as a constructor argument. Not all basis functions need to have this.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+        self._max_exactness = max_exactness
+        self._points = []
+        self._levels = []
+        self._update()
```
mphoward, line 33: I don't think you should call `_update` here, this would require any deriving class to have finished its initialization before the parent constructor is called. In fact, I would guess `_update` is not required if there is no `max_exactness`.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+    def points(self):
+        """Extrema assigned to Smolyak indices"""
+        return self._points
+
+    @property
+    def levels(self):
+        """index of each point that belongs to each grid level"""
+        return self._levels
+
+    @max_exactness.setter
```
mphoward, line 50: Just FYI: setters should be grouped next to their property for ease of reading.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+    @abc.abstractmethod
+    def _update(self):
+        """Update properties described by level of exactness
+        Compute points of basis function and the Smolyak index of
+        each grid level between 0 and max_exactness + 1 
+        """
+        pass
+
+    @abc.abstractmethod
+    def basis(self,x,n):
```
mphoward, line 85: I think we should reverse the order of these arguments and convert to a call method
```suggestion
    def __call__(self, n, x):
```
so that it matches more like the math formatting `T_n(x)` like
```
T = SomeBasisFunction()
T(1, x)
```

In general, does every basis function also take `n` as an argument in order to call it? I feel like we might need to think a bit more about that.

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+            self._levels = self._levels[:self._max_exactness+1]
+
+
+    @abc.abstractmethod
+    def _update(self):
+        """Update properties described by level of exactness
+        Compute points of basis function and the Smolyak index of
+        each grid level between 0 and max_exactness + 1 
+        """
+        pass
```
mphoward, line 82: Yes, suggest to remove: not all basis functions need to have this method. Instead, you should leave it up to the deriving classes to decide if they need to call `_update` before they evaluate basis.
```suggestion
```

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+        T_1(x) = x
+        T_{n+1}(x) = 2xT_n(x) - T_{n-1}(x)
+
+    The extrema of the polynomials are calculated via the following
+    equation:
+
+    ..math:
+        x_{n,j}^* = -\cos\left((\frac{j-1}{n-1}\pi\right), j = 1,...,n
+    """
+
```
mphoward, line 121: Here you can add a constructor with an argument named like `max_n` indicating the maximum value of `n` allowed for the basis function (your previous `max_exactness`).

basis/basis.py
```diff
@@ -0,0 +1,172 @@
...
+        y : float
+            output
+        """
+        pass
+
+
+class ChebyshevFirstKind(BasisFunction):
+    r"""Basis Function of Chebyshev polynomials of the first kind
+    Chebyshev polynomials are a set of polynomials described by the
+    following recurrence relation:
```
mphoward, line 108: This documentation is a bit unclear to me: I think that this is actually a family of basis functions, right? (See comment above on whether `__call__` should include `n`)

tests/basis_test.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+from basis.basis import BasisFunction, ChebyshevFirstKind
+
+@pytest.fixture
+def expected_extrema_2():
```
mphoward, line 5: How were these generated? You should note the source, and if possible, write them using exact math rather than decimals.

tests/basis_test.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+from basis.basis import BasisFunction, ChebyshevFirstKind
```
mphoward, line 2:
```suggestion
from smolyay.basis import BasisFunction, ChebyshevFirstKind
```

mphoward (changes requested):
Good start Michaela! Here are some suggestions and comments.

tests/basis_test.py
mzbush: They were generated from the old code files that Mohammadreza and I made earlier in the summer. I can represent everything up except some of the extrema for exactness 4 in mathematical terms.

I have found that ones like 1/sqrt(2) are not floating point precise with the answer the code gives, but they are very close.

mphoward: Gotcha, I think it is probably sufficient to go to exactness of 3 so you can write them all down “exactly”. Then, your unit tests should use floating point comparison with a tolerance.

basis/basis.py
mphoward: I thought more about this, and I think it is OK to keep n because it corresponds to one of the “point” indexes, right?

basis/basis.py
mzbush: max_n isn't same as max_exactness. 
Would it be sufficient to add an equation specifying how `max_n` and `max_exactness` are related? 
`max_n = 2**(max_exactness-1)+1`

mphoward: Maybe `num_levels`? Since based on what value you choose here, there will be a certain number of levels of points, right?

basis/basis.py
mzbush: It refers to the degree of the Chebyshev polynomial being called. In the BasisFunction class, I operated under the assumption that each basis function would be a series of related functions that are distinguished from each other by their degree like the Chebyshev polynomials or some variable analogous to a degree.

For the Chebyshev polynomials in particular, they do correspond to the point indexes. To construct the big matrix used to calculate the coefficients from slide 10 on the presentation, you have to feed the values in grid_point_indexes as n and the values of points[grid_point_indexes] as the x.

mphoward: OK, so then is “n” referring to a level or a point? It seems like a level because one Chebyshev polynomial adds multiple points, right?

basis/basis.py
mzbush: Yes. `max_exactness` is supposed to be the number of grid levels described in `self._levels` minus 1. That's its relationship with the number of grid levels. I called it `max_exactness` because it's meant to be exactly equal to the exactness parameter in `IndexGrid`, and I feel it's important if that relationship is very clear in the variable names.

Do you want `max_exactness` to be renamed `num_levels` and they have the same value? Should `num_levels` be `max_exactness` + 1?

mphoward: Let’s discuss together at our next meeting. We planned to generalize the IndexGrid when you weren’t there, and so not all IndexGrids will have an exactness variable.

basis/basis.py
mzbush: Neither really when it's used. 
The points added from one grid level are the extrema of the Chebyshev polynomial with degree `n = 2**(grid_level) + 1`
Each point then gets a Smolyak index from 0 to 2**(grid_level) + 1, and the Smolyak index of each point is reflected by its location in `points`. 
Then those Smolyak indexes get associated with a Chebyshev polynomial that has a degree equal to the Smolyak index when the surrogate function terms are being constructed. grid_point_indexes tells you the degree of the Chebyshev polynomial for each surrogate function terms.

mphoward: OK, I see, thanks! Let’s add this to the discussion list then.

smolyay/basis/basis.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @property
+    def sample_flag(self):
+        """list of what objects in basis_set to sample points from"""
+        return self._sample_flag
+
+    @sample_flag.setter
+    def sample_flag(self,sample_flag):
+        self._sample_flag = sample_flag
+        self._need_update = True
```
mphoward, line 184: I don't quite follow the purpose/usage of this, could you try explain more?

smolyay/basis/basis.py
```diff
@@ -0,0 +1,249 @@
...
+        super().__init__(sample_flag,basis_function)
+        self._update()
+
+    def _update(self):
+        """Update basis_set and get points"""
+        self._basis_set = []
+        self._all_points = []
+
+        for i in range(0,len(self._sample_flag)):
+            self._basis_set.append(self._basis_function(i))
```
mphoward, line 223: I also don't quite follow this, couldn't this be passed to the constructor as a list of concrete objects, rather than the class to be initialized?

mphoward (changes requested):
Thanks Michaela! This is moving in a good direction. I have some questions related to the design of the basis function sets, if you could please clarify there.

smolyay/basis/basis.py
```diff
@@ -0,0 +1,249 @@
...
+            self._n = n
+            self._points = []
+
+    def _compute_points(self):
+        """Compute extrema of Chebyshev polynomial of the first kind"""
+        if self._n == 0:
+            self._points = [0]
+        else:
+            for i in range(0,self._n+1):
+                temp = round(-math.cos(math.pi*(i)/(self._n)),15) + 0
```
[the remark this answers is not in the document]
mzbush: The rounding is to ensure that anything close to zero like 1e-16 is rounded down to zero and values like 0.9999999999 is seen as 1. The + 0 part causes -0 and +0 float values to be read as a plain 0. 
This ensures that the code correctly recognizes duplicate extrema values as duplicates.

smolyay/basis/basis.py
```diff
@@ -0,0 +1,249 @@
...
+    ``basis_function`` is the BasisFunction child class that populates
+    basis_set
+    ''all_points'' 1D points taken from the BasisFunction objects specified
+    by sample_flag
+    ``need_update`` is a boolean parameter that specifies when properties
+    of the class need to be recomputed
+
+    """
+
+    def __init__(self,sample_flag,basis_function):
```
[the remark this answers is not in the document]
mzbush: How basis_set is supposed to be constructed is something for the class to figure out based on it's inputs.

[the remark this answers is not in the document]
mphoward: Please see my comments below related to the use of `sample_flag` and constructing the basis set.

smolyay/basis/basis.py
mzbush: The points that BasisFunctionSet is supposed to get are taken from specific basis functions. In the case of the Chebyshev polynomial, it takes the extrema of polynomials of degree 0, 2, 4, 8, 16, and so on. `basis_set` stores a list of objects from a chebyshev polynomial of degree 0 to the max polynomial degree its supposed to sample extrema from. 

However, BasisFunctionSet has no way of knowing which objects in` basis_set` need to be sampled, so there needs to be a variable that tells the class that information. 

`sample_flag = [True, False, True, False, True, False, False, False, True]' means sample points from the objects in 'basis_set' at indexes 0, 2, 4, and 8. Because of this setup, 'basis_set' will always be the same length as 'sample_flag'.

I've been talking to Mohammadreza and think that it would be a good idea to replace sample_flag with a list of the indexes to sample points from. Something like `sample_index = [0,2,4,8]` We figured we should first hold it off until discussing it in our next meeting.

smolyay/basis/basis.py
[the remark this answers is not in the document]
mphoward: Ah, so that should be the job of whatever code detects duplicates, and not the basis function itself. I would recommend you turn this into the normal math (no rounding), then use `numpy.isclose` with some specified tolerance to detect duplicates when this is actually needed.

smolyay/basis/basis.py
mphoward: Ah OK, I see. I definitely like the second solution, it is easier to understand. I agree we should talk about how the `BasisFunctionSet` is going to be used subsequently before deciding on a specific interface. There may be some other options in addition to these two.

smolyay/basis/basis.py
mzbush: Technically, yes. basis_set could be an input for the constructor. However, given that there was only one child of BasisFunction, I didn't see the need to make creating basis_set something that needed to be calculated beforehand.

I can have BasisFunctionSet require basis_set as an input instead if that would be better.

mphoward: One advantage of doing that would be that `sample_indexes` (or whatever it is called) would refer to the index of the basis function in this list, rather than a constructor argument.

mphoward: Actually: this is going to be important. The general `BasisFunction` does not take `n` as an input, so this line could actually fail for other potential children.

mzbush: Well, the fact that it assumed that n was an input needed for `basis_function` is why it's called RecurrenceSet. It assumed that `basis_function` can be described by a recurrence relation and thus the objects need to know where they are in that recurrence relation.

mphoward: You would need to have a new type of `BasisFunction` to enforce this type check, but I think it is probably not a good idea. You don't want to force a class to have a specific constructor, and this would mean the class can only have one argument (`n`) in its constructor.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+    following recurrence relation:
+
+    ..math:
+        T_0(x) = 1
+        T_1(x) = x
+        T_{n+1}(x) = 2xT_n(x) - T_{n-1}(x)
+
+    ``n`` is the degree of the polynomial this object describes
+    ``points`` represents the extrema of the polynomial of this degree on
+    the domain [-1,1]
```
mphoward, line 56: Update this once you replace `points` by `roots` and `extrema` in the base class.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    The extrema of the polynomial are calculated via the following
+    equation:
+
+    ..math:
+        x_{n,j}^* = -\cos\left((\frac{j-1}{n-1}\pi\right), j = 1,...,n
+        n = 2^{exactness-1}+1
```
mphoward, line 63: This formula holds for any `n` right? With the new layout, there is no concept of exactness.
```suggestion
        x_i^* = -\cos\left((\frac{i-1}{n-1}\pi\right), i = 1,...,n
```

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Parameters
+    ----------
+    n : int
+        degree of the Chebyshev polynomial
```
mphoward, line 68:
```suggestion
        Degree of the polynomial.
```

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+    r"""Basis function Chebyshev polynomials of the first kind
+    Chebyshev polynomials are a sequence of polynomials described by the
+    following recurrence relation:
+
+    ..math:
+        T_0(x) = 1
+        T_1(x) = x
+        T_{n+1}(x) = 2xT_n(x) - T_{n-1}(x)
+
+    ``n`` is the degree of the polynomial this object describes
```
mphoward, line 54: I think you need to be clear here that this is only _one_ term (the `n`-th term) in the polynomial, and not the polynomial of degree `n`. This is important because it is different from the NumPy Chebyshev polynomial:
https://numpy.org/doc/stable/reference/routines.polynomials.chebyshev.html

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+        """degree of polynomial"""
+        return self._n
+
+    @n.setter
+    def n(self,n):
+        if self._n != n:
+            self._n = n
+            self._points = []
+
+    def _compute_points(self):
```
mphoward, line 93: Is this function equivalent to:
https://numpy.org/doc/stable/reference/generated/numpy.polynomial.chebyshev.chebpts1.html#numpy.polynomial.chebyshev.chebpts1
?

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+    @property
+    def n(self):
+        """degree of polynomial"""
+        return self._n
+
+    @n.setter
+    def n(self,n):
+        if self._n != n:
+            self._n = n
+            self._points = []
```
mphoward, line 91: Is there a point to being able to change the value of `n`? I would probably make this read-only.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _compute_points(self):
+        """Compute extrema of Chebyshev polynomial of the first kind"""
+        if self._n == 0:
+            self._points = [0]
+        else:
+            i = numpy.arange(self._n+1)
+            self._points = list(-numpy.cos(numpy.pi*i/self._n))
+
+    def __call__(self,x):
```
mphoward, line 101: Just making a note that we may want to think about how this calculation is done later. Mohammadreza found a method using branching, and I read that NumPy evaluates Chebyshev polynomials using something called Clenshaw recursion:
https://numpy.org/doc/stable/reference/generated/numpy.polynomial.chebyshev.chebval.html#numpy.polynomial.chebyshev.chebval
https://en.wikipedia.org/wiki/Clenshaw_algorithm
I don't know if the later helps here or not, since we have only one term in the series.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+class BasisFunctionSet():
+    """Set of basis functions
+    Creates a framework for a set of basis functions to be used as the
+    building blocks for a surrogate model.
+    ``basis_set`` is a list of BasisFunction objects
+    ''all_points'' 1D points taken from the BasisFunction objects
+
+    """
+
+    def __init__(self,all_points,basis_set):
```
mphoward, line 137: Point of discussion: do we want to require the user to split the `points` and `basis` into separate lists when they construct this, like `(p0,p1,...),(b0,b1,...)`? An alternative would be to require they come in tuples like: `((p0,b0), (p1,b1), ...)`. If we go with the first option, you need to error check that the arrays have the same length.

In any case, see suggestions below for renaming these variables.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+        self._all_points = all_points
+
+    @property
+    def all_points(self):
+        """All points for basis function set at some level of precision"""
+        return self._all_points
+
+    @all_points.setter
+    def all_points(self,all_points):
+        self._all_points = all_points
```
mphoward, line 148: I would name this `points` and make it readonly. (You cannot change one of these properties without changing all of them.) `all_` does not give any new information here

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @property
+    def basis_set(self):
+        """list of BasisFunction objects"""
+        return self._basis_set
+
+    @basis_set.setter
+    def basis_set(self,basis_set):
+        self._basis_set = basis_set
```
mphoward, line 157: I would rename this `basis_functions` and make it readonly. The class is the `BasisFunctionSet`, so an attribute `basis_set` doesn't clarify what the attribute actually is

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+    def levels(self):
+        """list of list that show the points for each grid level"""
+        return self._levels
+
+    @levels.setter
+    def levels(self,levels):
+        self._levels = levels
+
+
+def make_nested_chebyshev_points(exactness,basis_function):
```
mphoward, line 185: This is nice! I would make it a classmethod of the Chebyshev basis function called `make_nested_set` that takes only one argument `exactness`, currently.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    levels = [[0]]
+    basis_set = []
+    all_points = [0]
+    max_degree = 2**exactness
+    if exactness == 0:
+        max_degree = 0
+    start_level_index = 1
+    for i in range(0,max_degree+1):
+        basis_set.append(basis_function(i))
```
mphoward, line 211: This can be done in one line!
```suggestion
    basis_set = [basis_function(n) for n in range(max_degree+1)]
```

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+        data structure for points
+    """
+
+    levels = [[0]]
+    basis_set = []
+    all_points = [0]
+    max_degree = 2**exactness
+    if exactness == 0:
+        max_degree = 0
+    start_level_index = 1
```
mphoward, line 209: This pre-initialization stuff is confusing. Shouldn't the `ChebyshevFirstKind(0)` be able to return the right point `0` so there is no special case?

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+    for j in range(1,exactness+1):
+        degree_sample = 2**j
+        new_points = basis_set[degree_sample].points
+        end_level_index = start_level_index
+        for k in range(0,len(new_points)):
+            if not numpy.isclose(all_points,new_points[k]).any():
+                all_points.append(new_points[k])
+                end_level_index = end_level_index + 1
+        levels.append(list(range(start_level_index,end_level_index)))
+        start_level_index = end_level_index
```
mphoward, line 222: I think some of this loop counting can be condensed, but I'd like to think about whether it's really necessary to treat `j == 0` as a special case before commenting on this.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    nested_set = NestedBasisFunctionSet(all_points,basis_set,levels)
+    return nested_set
```
mphoward, line 225:
```suggestion
    return NestedBasisFunctionSet(all_points,basis_set,levels)
```

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    """
+
+    def __init__(self):
+        self._points = []
+
+    @abc.abstractproperty
+    def points(self):
+        """The points of the basis function assigned to Smolyak indices"""
+        pass
```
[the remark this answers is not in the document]
mzbush: The Chebyshev polynomial of the 1st kind can use both its extrema in addition to its root because its extrema within the domain [-1,1] lie in the range [-1,1]. The Chebyshev polynomial of the 2nd kind's extrema for that domain don't lie in that range which makes them not very useful. Having both be useful seems pretty unique and I don't think that distinction makes a lot of sense for many basis function besides Chebyshev of the 1st kind.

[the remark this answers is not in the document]
mphoward: OK, that's fair. We can just be clear in the documentation then, once you add the ability to choose how `points` is defined.

smolyay/basis.py
mzbush: Not exactly. That function appears to print the new extrema added by every nth term. However, it doesn't include -1 or 1. 
I could modify _compute_points to use this. The check in make_nested_chebyshev_points line 217 wouldn't have to be done. However, since it doesn't include -1 and 1, I would need to make special considerations in _compute_points and make_nested_chebyshev_points to make sure they're added correctly.

smolyay/basis.py
mzbush: It's more that the terms in the Chebyshev sequence being sampled are 0, 2, 4, 8, and so on for 2^n. The problem is that 2^0 is 1 and not 0 and I didn't want to sample from `ChebyshevFirstKind(1)` because it's points are [-1,1]. So I just did the initial part manually because I didn't think of an easy way to modify `degree_sample = 2**j` so that `degree_sample` = 0 when `j` = 0
I will note that `numpy.polynomial.chebyshev.chebpts1(1)` does give 0. However, `numpy.polynomial.chebyshev.chebpts1(2)` is [-0.7071067811865476, 0.7071067811865476] and those points come from the 4th term in the Chebyshev sequence, not the 2nd. This is part of what makes potentially using them in `compute_points `difficult.

smolyay/basis.py
```diff
@@ -0,0 +1,226 @@
...
+import numpy
+
+class BasisFunction(abc.ABC):
+    """Abstract class for basis functions.
+
+    Creates a framework for basis functions that are the building blocks
+    for the terms in the surrogate model. 
+    ``points`` property represents all the unique 1D points of the 
+    basis function associated with a Smolyak index described by the 
+    class IndexGrid.
```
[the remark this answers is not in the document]
mzbush: I've changed the documentation, I'm not sure where to go from here to improve it.

smolyay/basis.py
mzbush: I think having separate lists would be easier to implement. I can include error checking.

smolyay/basis.py
mzbush: I'm not sure what you mean by "classmethod of the Chebyshev basis function" but I can do the rest.

mphoward: I mean this function should be a classmethod for ChebyshevFirstKind:

```
@classmethod
def make_nested_set(cls, exactness):
    …
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+    Object describes the nth term in the set of polynomials as designated
+    by the property ``n``
+    ``points`` represents the extrema of the polynomial of this degree on
+    the domain [-1,1]
+
+    The extrema of the polynomial are calculated via the following
+    equation:
+
+    ..math:
+        x_{i}^* = -\cos\left((\frac{i-1}{n-1}\pi\right), i = 1,...,n
```
mphoward, line 62:
```suggestion
    r"""Chebyshev polynomial of the first kind.

    The Chebyshev polynomial of degree *n* is defined by the
    recursive relationship:

    ..math::

        T_0(x) = 1
        T_1(x) = x
        T_{n+1}(x) = 2x T_n(x) - T_{n-1}(x)

    The :attr:`points` for this polynomial are the extrema on the domain
    :math:`[-1,1]`:
   
    ..math::

        x_i^* = -\cos(\pi i/n), i = 0,...,n
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Parameters
+    ----------
+    n : int
+        Degree of the Chebyshev polynomial.
+    """
+
+    def __init__(self,n):
+        super().__init__()
+        self._n = n
```
mphoward, line 72: Since `n` is readonly and the formula for computing the points is simple, just take care of it in the constructor:
```suggestion
        self._n = n
        self._points = list(numpy.polynomial.chebyshev.chebpts2(n+1)) if n > 0 else [0]
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+
+    @property
+    def points(self):
+        """extrema of polynomial"""
+        if len(self._points) == 0:
+            self._compute_points()
```
mphoward, line 78:
```suggestion
        """list: Sampling points at extrema of polynomial."""
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+    def n(self):
+        """degree of polynomial"""
+        return self._n
+
+    def _compute_points(self):
+        """Compute extrema of Chebyshev polynomial of the first kind"""
+        if self._n == 0:
+            self._points = [0]
+        else:
+            self._points = list(numpy.polynomial.chebyshev.chebpts2(self._n+1))
```
mphoward, line 91:
```suggestion
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+        Parameters
+        ----------
+        x : float
+            Input arguement.
+
+        Returns
+        -------
+        float
+            Value of Chebyshev polynomial of the first kind.
+        """
```
mphoward, line 107:
```suggestion
        r"""Evaluate the basis function.

        The Chebyshev polynomial is evaluated using the combinatorial formula:
        
        .. math::
        
            T_n(x) = \sum_{k=0}^{\lfloor n/2 \rfloor} {n \ choose 2k} (x^2-1)^k x^{n-2k}

        for :math:`n \ge 2`.

        Parameters
        ----------
        x : float
            One-dimensional point on :math:`[-1,1]`.

        Returns
        -------
        float
            Value of Chebyshev polynomial of the first kind.

        """
```

smolyay/basis.py
```diff
@@ -0,0 +1,205 @@
...
+        Parameters
+        ----------
+        exactness : int
+            Level of exactness to calculate points to.
+
+        Returns
+        -------
+        NestedBasisFunctionSet object
+            Data structure for points.
+        """
```
mphoward, line 134:
```suggestion
        """Create a nested set of Chebyshev polynomials.

        A nested set is created up to a given level of ``exactness``,
        which corresponds to a highest-order Chebyshev polynomial of
        degree ``n = 2**exactness``.

        Each nesting level corresponds to the increasing powers of 2 going up to
        ``2**exactness``, with the first level being a special case. The generating
        Chebyshev polynomials are hence of degree (0, 2, 4, 8, ...). Each new point
        added in a level is paired with a basis function of increasing order.

        For example, for an ``exactness`` of 3, the generating polynomials are
        of degree 0, 2, 4, and 8, at each of 4 levels. There are 1, 2, 2, and 4
        new points added at each level. The polynomial at level 0 is of degree 0,
        the polynomials at level 1 are of degrees 1 and 2, those at level 2 are of
        degree 3 and 4, and those at level 3 are of degrees 5, 6, 7, and 8.

        Parameters
        ----------
        exactness : int
            Level of exactness.

        Returns
        -------
        NestedBasisFunctionSet
            Nested Chebyshev polynomials of the first kind.

        """
```

mphoward (changes requested):
Good job Michaela! Here are some suggestions (many of which you can apply directly, possibly just through the web interface) and one or two spots of documentation that need expansion. I will review the unit tests after this code is revised, but this looks like it is almost done from my end!

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+def test_set_nested_change_levels():
+    """Check NestedBasisFunctionSet updates points when level changes"""
+    levels = [[0]]
+    points = [0]
+    test_class = NestedBasisFunctionSet(points,[ChebyshevFirstKind(0)],levels)
+    test_class.levels = [[0],[1,2],[3,4]]
+    assert test_class.levels == [[0],[1,2],[3,4]]
+
+def test_set_compute_function(expected_points_3_set):
```
mphoward, line 141: Same thing here.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+    """Chebyshev polynomial degree 1 should return input"""
+    test_class = ChebyshevFirstKind(1)
+    for i in range(0,16):
+        assert test_class(i) == i
+
+def test_basis_input_1():
+    """Chebyshev polynomial input 1 returns 1 for any degree n"""
+    for i in range(0,16):
+        test_class = ChebyshevFirstKind(i)
+        assert test_class(1) == 1
```
mphoward, line 76: I think this test is not necessary if you include `-1` and `1` as inputs.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+@pytest.mark.parametrize("x, n, expected",[(2,6,1351),(-3,9,-3880899),
+    (13,4,227137),(8,3,2024),(4,9,58106404),(14,5,8550374),(0,12,1),
+    (7,5,262087),(-2,4,97),(3,9,3880899),(9,4,51841),(2,11,978122),
+    (4,7,937444),(-3,2,17),(6,4,10081),(2,8,18817)])
+def test_basis_random_points(x,n,expected):
+    """Test chebyshev polynomial at some degree at some input"""
+    test_class = ChebyshevFirstKind(n)
+    assert test_class(x) == expected
```
mphoward, line 85: Rather than run at this random set of points, maybe you could adapt the use of parametrize to just be over the value of `n`, then use `numpy`/`scipy` to compute the value of the polynomial at the points you want by a different route? That would probably be more comprehensive.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+def test_is_basis_abstract():
+    """Check BasisFunction is an abstract class"""
+    with pytest.raises(TypeError):
+        test_class = BasisFunction()
```
mphoward, line 90: This test is not necessary, suggest to delete.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+    assert test_class.points == [0]
+    assert test_class.levels == [[0]]
+
+def test_set_nested_change_levels():
+    """Check NestedBasisFunctionSet updates points when level changes"""
+    levels = [[0]]
+    points = [0]
+    test_class = NestedBasisFunctionSet(points,[ChebyshevFirstKind(0)],levels)
+    test_class.levels = [[0],[1,2],[3,4]]
+    assert test_class.levels == [[0],[1,2],[3,4]]
```
mphoward, line 139: This test would produce an unusable `NestedBasisFunctionSet`, which actually points out something you should go back and look at. The indexes that are used in `levels` must all be smaller than the length of the points/basis functions. Otherwise, you potentially get an index error. You should fix this in the main code, then write some tests that both test for the error and also will work correctly.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+    test_class = ChebyshevFirstKind.make_nested_set(3)
+    assert numpy.allclose(
+            test_class.points,expected_points_3_set,atol=1e-10)
+    assert test_class.levels == [[0],[1,2],[3,4],[5,6,7,8]]
+    basis_functions = test_class.basis_functions
+    assert len(basis_functions) == 9
+    for i in range(0,len(basis_functions)):
+        assert basis_functions[i].n == i
+
+def test_set_compute_empty():
```
mphoward, line 152: The names of these tests are confusing. Can you maybe add `cheb` like you did above, and also include `make_nested` or some variation of that?

mphoward (changes requested):
Thanks for making the other changes! Here are some comments on the tests.

tests/basis_test.py
```diff
@@ -0,0 +1,161 @@
...
+    """test initial when degree is 8"""
+    test_class = ChebyshevFirstKind(8)
+    assert test_class.n == 8
+    assert numpy.allclose(test_class.points,expected_points_8,atol=1e-10)
+
+def test_basis_degree_0():
+    """Chebyshev polynomial degree 0 is 1 for all inputs"""
+    test_class = ChebyshevFirstKind(0)
+    for i in range(0,16):
+        assert test_class(i) == 1
```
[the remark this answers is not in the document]
mzbush: No, because what is being tested is not the points but the polynomial. T_0(x) = 1 for all x and T_1(x) = x for all x. The inputs for the call method are not expected to always be on the range [-1,1].

smolyay/basis.py
```diff
@@ -0,0 +1,232 @@
...
+        if len(self._points) == 0:
+            self._compute_points()
+        return self._points
+
+    @property
+    def n(self):
+        """degree of polynomial"""
+        return self._n
+
+    def __call__(self,x):
```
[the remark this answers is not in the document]
mzbush: No, because the inputs for the call method are not expected to always be on the range [-1,1]. For example, If the input space for the real functions parameters aren't on that range (like on the branin function the surrogate was matching it on the range of [-5,10] for x1 and [-5,15] for x2).

tests/basis_test.py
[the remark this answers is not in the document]
mphoward: No, this is incorrect. These polynomials are typically defined on [-1,1]. For example, x does not have extrema at -1 and 1 if T_1 is not defined on this domain. We have also documented that the basis functions are defined on this domain (see docstring).

smolyay/basis.py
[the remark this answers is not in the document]
mphoward: See comment above. It is the responsibility of other classes to shift and scale their inputs to the relevant domain. Your and Mohammadreza’s code did this before.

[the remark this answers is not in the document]
mzbush: I am looking at my code that I made at the beginning of the summer and the Chebyshev polynomial is definitely called outside of that domain. Mohammadreza's does the same thing.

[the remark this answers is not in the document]
mphoward: Then that code was incorrect. The real function may have a different domain, but the points/inputs need to be scaled from/to [-1,1] before the Chebyshev polynomial is called. I believe Dr. K explained this to us?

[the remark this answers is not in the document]
mzbush: Yeah, that appears to be correct with the literature.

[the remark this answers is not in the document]
cakieslich: It seems like this is resolved, but yes we typically have two versions of the grid points. One on the domain of [-1 1] in all dimensions, which is used to sample the Smolyak polynomial. The other version of the grid points are those where the [-1 1] domain is transformed to that of the problem at hand, which is used to sample the model that is being approximated.

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+        r"""Evaluate the basis function.
+
+        The Chebyshev polynomial is evaluated using the combinatorial formula:
+        
+        .. math::
+        
+            T_n(x) = \sum_{k=0}^{\lfloor n/2 \rfloor} {n \ choose 2k} (x^2-1)^k x^{n-2k}
```
mphoward, line 93:
```suggestion
            T_n(x) = \sum_{k=0}^{\lfloor n/2 \rfloor} {n \choose 2k} (x^2-1)^k x^{n-2k}
```

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+
+        for :math:`n \ge 2`.
```
mphoward, line 95:
```suggestion
        for :math:`n \ge 2`, and by the direct formula for the other values of *n*.
```

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+    """Nested set of basis functions and points.
+
+    Nested points/basis function grow in levels, such that an approximation
+    of a given level uses not only its sampling points but also all the points 
+    at lower levels. Nested sets (similarly to ogres) are like onions.
+
+    Parameters
+    ---------
+    levels : list of lists
+        determines which level each point is added on
```
mphoward, line 223:
```suggestion
       Assignment of points/basis functions to each level.
```

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+        return self._levels
+
+    @levels.setter
+    def levels(self,levels):
+        """
+        Raises
+        ------
+        IndexError
+            max index must be less than total number of points.
+        """
```
mphoward, line 242: I think you will need to document this with the property getter (above), as the getter and setter are probably not both rendered by sphinx. Please move.

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+        if any([any(numpy.greater(y,len(self._points))) for y in levels]):
```
mphoward, line 243: This would be an easier/clearer way to do this:
```suggestion
        if numpy.any(numpy.concatenate(levels) > len(self.points)):
```

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+            raise IndexError("max level index must be less than total "
+                    "number of points.")
+        else:
+            self._levels = levels
```
mphoward, line 247: Usually, you don't do `if-else` here, since if the exception is raised the code stops
```suggestion
        self._levels = levels
```

tests/basis_test.py
```diff
@@ -0,0 +1,157 @@
...
+    return [0, -1.0, 1.0, -1/numpy.sqrt(2), 1/numpy.sqrt(2),
+            -numpy.sqrt(numpy.sqrt(2)+1)/(2**0.75),
+            -numpy.sqrt(numpy.sqrt(2)-1)/(2**0.75),
+            numpy.sqrt(numpy.sqrt(2)-1)/(2**0.75),
+            numpy.sqrt(numpy.sqrt(2)+1)/(2**0.75)]
+
+def test_cheb_initial_zero():
+    """test degree of zero"""
+    f = ChebyshevFirstKind(0)
+    assert f.n == 0
```
mphoward, line 20: Check `f.points` here?

tests/basis_test.py
```diff
@@ -0,0 +1,157 @@
...
+    """Chebyshev polynomial degree 0 is 1 for all inputs"""
+    f = ChebyshevFirstKind(0)
+    for i in [-1, -0.5, 0, 0.5, 1]:
+        assert f(i) == 1
+
+def test_cheb_call_degree_1():
+    """Chebyshev polynomial degree 1 should return input"""
+    f = ChebyshevFirstKind(1)
+    for i in [-1, -0.5, 0, 0.5, 1]:
+        assert f(i) == i
```
mphoward, line 61: I would fuse these with the tests above, since they are just one more assert after checking the points

tests/basis_test.py
```diff
@@ -0,0 +1,157 @@
...
+
+def test_cheb_call_random_points():
+    """Test chebyshev polynomial at some degree at some input"""
+    numpy.random.seed(567)
+    n = numpy.random.randint(20,size = 20)
+    x = numpy.random.rand(20) * 2 - 1
+    for i in range(20):
+        f = ChebyshevFirstKind(n[i])
+        assert numpy.isclose(f(x[i]),special.eval_chebyt(n[i],x[i]))
```
mphoward, line 70: This would be a bit more pythonic:
```suggestion
    ns = numpy.random.randint(20,size = 20)
    xs = numpy.random.rand(20) * 2 - 1
    for n,x in zip(ns,xs):
        f = ChebyshevFirstKind(n)
        assert numpy.isclose(f(x),special.eval_chebyt(n,x))
```

mphoward (approved):
Nice job Michaela! A couple last minor things, but then this is good to go. I don't need to see it again after you implement these changes.

mphoward (approved):

smolyay/basis.py
```diff
@@ -0,0 +1,248 @@
...
+        T_{n+1}(x) = 2x T_n(x) - T_{n-1}(x)
+
+    The :attr:`points` for this polynomial are the extrema on the domain
+    :math:`[-1,1]`:
+   
+    .. math::
+
+        x_i^* = -\cos(\pi i/n), i = 0,...,n
+
+For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
```
mphoward, line 64: Please fix this indentation.

cakieslich (approved):

PR #3, opened 2022-07-19
Feature/surrogate

mzf0069:
"SurrogateFunction" class constructs a surrogate to approximate a real function based on the basis function chosen for the surrogate and level of approximation.


smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+        self._levels = self._nested_basis_set.levels
+
+    @property
+    def nested_basis_set(self):
+        """:class:`NestedBasisSet`: Nested basis set (data structure)."""
+        return self._nested_basis_set
+
+    @nested_basis_set.setter
+    def nested_basis_set(self, nested_basis_set):
+        self._nested_basis_set = nested_basis_set
```
mphoward, line 135: You can remove these if you put them in the parent class.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+
+    def indices(self, dimension, expand_indicy=[], drop_indicy=[]):
```
mphoward, line 137: We talked about removing the expand/drop here. Can you make that simplification before review?

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+
+        Update the grid points (and their integers), basis functions and
+        indicies.
+
+        Parameters
+        ----------
+        dimension: int
+            Number of independent variables.
+        """
+        self._dimension = dimension
```
mphoward, line 221: Dimension should not be an attribute, it is only used during the `__call__`.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+        -------
+        :class:`SmolyakGrid`
+            Grid points (and their indexes),their corresponding functions and
+            indices.
+        """
+        self._dimension = dimension
+        self._update(self._dimension)
+
+        return SmolyakGrid(self._grid_points_indexes, self._grid_points,
+                           self._grid_points_basis, self._indices)
```
mphoward, line 208: Similarly here, you should not be saving these into attributes. Get rid of `_update`, and put those implementation parts here. These should be local variables that you use to return this grid.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+        self._grid_points_indexes = list(itertools.product(
+            *[points_indexes for point in range(self._dimension)]))
+        self._grid_points_indexes = list(map(list, self._grid_points_indexes))
+        self._grid_points = numpy.array(self._points
+                                        )[numpy.array(self._grid_points_indexes
+                                                      )].tolist()
+        self._grid_points_basis = (numpy.array(self._basis)[
+            numpy.array(self._grid_points_indexes)].tolist())
+        return IndexGrid(self._grid_points_indexes, self._grid_points,
+                         self._grid_points_basis)
```
mphoward, line 304: Same here, no need to store these in attributes.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+    """Grid points for interpolation.
+
+    Depending on the approach and dimensionality
+    , a set of grid points, and their corresponding
+    basis functions are generated. These grid points
+    and functions can then be then used for approximation
+    of complex systems.
+    """
+
+    def __init__(self):
```
mphoward, line 20: I think you should add a `basis_set` argument here, save it to `self._basis_set`, then expose it as the `basis_set` property. All classes are likely to use it.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+    Parameters
+    ----------
+    basis_set: :class:`BasisFunctionSet`
+        points, basis functions and levels.
+    """
+
+    def __init__(self, basis_set):
+        self._basis_set = basis_set
+        self._points = self._basis_set.points
+        self._basis = self._basis_set.basis_functions
```
mphoward, line 272: It would be better to just access these from `self._basis_set`, as it is duplicative to copy them.

smolyay/grid.py
```diff
@@ -0,0 +1,458 @@
...
+        self._grid_points = numpy.array(self._points
+                                        )[numpy.array(self._grid_points_indexes
+                                                      )].tolist()
+        self._grid_points_basis = (numpy.array(self._basis)[
+            numpy.array(self._grid_points_indexes)].tolist())
+        return IndexGrid(self._grid_points_indexes, self._grid_points,
+                         self._grid_points_basis)
+
+
+class IndexGrid():
```
mphoward, line 307:
```suggestion
class IndexGrid:
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,540 @@
...
+
+        # (C)
+        h += 1
+        t = r[h-1]
+        r[h-1] = 0
+        r[0] = t-1
+        r[h] += 1
+        yield list(r) if include_zero else (numpy.array(r)+1).tolist()
+
+########################################################################
```
mphoward, line 233: I think this is duplicated from above? If so, remove.

mphoward (changes requested):

smolyay/grid.py
mzf0069: The point of this method was first storing the levels (which is also in data structure), and expanding or dropping levels. Since, we are simplifying it for now, I don't see any reason to keep this.

smolyay/surrogate.py
mphoward: Why do you need it here if it is also in `grid.py`? If the code is the same, it should only be in one spot.

smolyay/grid.py
```diff
@@ -0,0 +1,362 @@
+
+import abc
+import numpy
+from smolyay.basis import (BasisFunction, ChebyshevFirstKind,
+                           BasisFunctionSet, NestedBasisFunctionSet)
+import itertools
```
mphoward, line 6: Per PEP8, import order should be core libraries, third party libraries, then your application, with a line break between these groups.
```suggestion
import abc
import itertools

import numpy

from smolyay.basis import (BasisFunction, ChebyshevFirstKind,
                           BasisFunctionSet, NestedBasisFunctionSet)
```

smolyay/grid.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Depending on the approach and dimensionality
+    , a set of grid points, and their corresponding
+    basis functions are generated. These grid points
+    and functions can then be then used for approximation
+    of complex systems.
+    """
+
+    def __init__(self, basis_set):
+        self._basis_set = basis_set
```
mphoward, line 20: If you do it this way, it will trigger the setter (in case there is anything special you want the setter to do)
```suggestion
        self.basis_set = basis_set
```

smolyay/grid.py
```diff
@@ -0,0 +1,362 @@
...
+        (:class:`ChebyshevFunction`, :class:`ChebyshevFunction`),
+        (:class:`ChebyshevFunction`, :class:`ChebyshevFunction`),
+        (:class:`ChebyshevFunction`, :class:`ChebyshevFunction`)]
+
+    """
+
+    def __init__(self, nested_basis_set):
+        super().__init__(nested_basis_set)
+        self._nested_basis_set = nested_basis_set
+        self._levels = self._nested_basis_set.levels
```
mphoward, line 131: The levels will be in the nested basis set, so you don't need it. Also, use the same notation as the parent constructor, but document the required type of the argument and error check it:
```suggestion
    def __init__(self, basis_set):
        super().__init__(basis_set)
        if not isinstance(self.basis_set, NestedBasisSet):
            raise TypeError('Basis set must be nested for Smolyak grid')
```

smolyay/grid.py
```diff
@@ -0,0 +1,362 @@
...
+        Returns
+        -------
+        :class:`SmolyakGrid`
+            Grid points (and their indexes),their corresponding functions and
+            levels.
+        """
+        grid_points_indexes = None
+        levels = []
+        for sum_of_levels in range(dimension,
+                                   dimension+len(self._levels)):
```
mphoward, line 155:
```suggestion
                                   dimension+len(self._basis_set.levels)):
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,469 @@
+import numpy
+
+
+class SmolyakGrid:
```
mphoward, line 4: Shouldn't this class be removed now that there's a generator in the `grid` module?

Similarly, is there anything in `smolyak.py` that hasn't been moved somewhere else already? I think `smolyak.py` can now be removed too.

mphoward (changes requested):

smolyay/grid.py
```diff
@@ -0,0 +1,396 @@
...
+                        (grid_points_indexes,
+                         grid_points_indexes_), axis=0)
+        # make integer grids, levels, grid points and basis functions
+        grid_points_indexes = grid_points_indexes.tolist()
+        level_lookup = {point: level for level, points in
+                        enumerate(self._basis_set.levels) for point in points}
+        levels = numpy.array([level_lookup[index] for index in
+                              numpy.concatenate(grid_points_indexes)]
+                             ).reshape((len(grid_points_indexes),
+                                        dimension)).tolist()
```
mphoward, line 214: Is this casting and reshaping necessary, or could it be done like:
```
levels = [[level_lookup[index] for index in indexes] for indexes in grid_point_indexes]
```
?

smolyay/grid.py
```diff
@@ -0,0 +1,396 @@
...
+        grid_points = numpy.array(self._basis_set.points
+                                  )[numpy.array(grid_points_indexes)].tolist()
+        grid_points_basis = numpy.array(self._basis_set.basis_functions)[
+            numpy.array(grid_points_indexes)].tolist()
```
mphoward, line 218: I'm a little nervous about using `numpy.array` here because the basis functions are objects. Could you instead do:
```
grid_points_basis = [[self._basis_set.basis_functions[index] for index in indexes] for indexes in grid_points_indexes]
```
as above?

Maybe it would actually make sense to do this all in a loop that appends to `levels`, `grid_points`, and `grid_points_basis` like:
```
levels = []
grid_points = []
grid_points_basis = []
for indexes in grid_points_indexes:
    levels.append([level_lookup[index] for index in indexes])
    grid_points.append([self._basis_set.points[index] for index in indexes])
    grid_points_basis.append([self._basis_set.basis_functions[index] for index in indexes])
```

smolyay/grid.py
```diff
@@ -0,0 +1,396 @@
...
+        points_indexes = numpy.arange(len(self._basis_set.points))
+        # make integer grids, grid points and basis functions
+        grid_points_indexes = list(itertools.product(
+            *[points_indexes for point in range(dimension)]))
+        grid_points_indexes = list(map(list, grid_points_indexes))
+        grid_points = numpy.array(self._basis_set.points
+                                  )[numpy.array(grid_points_indexes
+                                                )].tolist()
+        grid_points_basis = (numpy.array(self._basis_set.basis_functions)[
+            numpy.array(grid_points_indexes)].tolist())
```
mphoward, line 266: Same comment here about how these are generated.

smolyay/grid.py
```diff
@@ -0,0 +1,396 @@
...
+        Returns
+        -------
+        :class:`IndexGrid`
+            Grid points (and their indexes),their corresponding functions.
+        """
+        points_indexes = numpy.arange(len(self._basis_set.points))
+        # make integer grids, grid points and basis functions
+        grid_points_indexes = list(itertools.product(
+            *[points_indexes for point in range(dimension)]))
+        grid_points_indexes = list(map(list, grid_points_indexes))
```
mphoward, line 261: What is going on with list and map here? Is it because you need to make a list of lists?

smolyay/grid.py
```diff
@@ -0,0 +1,396 @@
...
+        """list: Grid points' indexes."""
+        return self._indexes
+
+    @property
+    def points(self):
+        """list: Grid points."""
+        return self._points
+
+    @property
+    def basis(self):
```
mphoward, line 300: Should you call this `basis_functions` for consistency with those sets?

tests/grid_test.py
```diff
@@ -0,0 +1,214 @@
...
+def test_error_basis_set_type():
+    """Test if raise TypeError works for IndexGridGenerator."""
+    basis_set = "not a BasisFunctionSet"
+    with pytest.raises(TypeError):
+        IndexGridGenerator(basis_set)
+
+
+def test_error_nested_basis_set_type():
+    """Test if raise TypeError works for SmolyakGridGenerator."""
+    basis_set = "not a NestedBasisFunctionSet"
```
mphoward, line 19: It would be a better test to use a regular `BasisFunctionSet` here rather than a string, as that would be a potential common mistake.

tests/grid_test.py
```diff
@@ -0,0 +1,214 @@
...
+
+def test_smolyak_levels_one_dimension():
+    """Test levels of the grid points. in one dimension."""
+    test_nested_basis_set = ChebyshevFirstKind.make_nested_set(3)
+    test_class = SmolyakGridGenerator(test_nested_basis_set)
+    expected_levels = [[0], [1], [1], [2], [2], [3], [3], [3], [3]]
+    assert test_class(1).levels == expected_levels
+
+
+def test_smolyak_levels():
```
mphoward, line 53: It might make sense to test the indexes, levels, etc. all in one test function, so you don't have to jump around as much reading. It's OK for tests to make multiple assertions.

mphoward (changes requested):
This is looking great! Here are a couple suggestions and requested changes.

smolyay/grid.py
mzf0069: It could be done. I was just trying to avoid having two for loops.

smolyay/grid.py
mzf0069: Same here. I was trying to avoid having two for loops.

smolyay/grid.py
mzf0069: Yes. the list(itertools.product) generates list of tuples.

smolyay/grid.py
mphoward: I think this is such a negligible performance thing, you should go with the append loop structure I suggested. It is more readable and avoids using NumPy arrays on an object

smolyay/grid.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        points_indexes = numpy.arange(len(self._basis_set.points))
+        # make integer grids, grid points and basis functions
+        grid_points_indexes = list(itertools.product(
+            *[points_indexes for point in range(dimension)]))
+        grid_points_indexes = list(map(list, grid_points_indexes))
+        grid_points = [[self._basis_set.points[index] for index in indexes] for
+                       indexes in grid_points_indexes]
+        grid_points_basis = [[self._basis_set.basis_functions[index] for index
+                              in indexes] for indexes in grid_points_indexes]
```
mphoward, line 265: I would suggest a similar change here as above.

mphoward (changes requested):
Just a couple small things and then I'm good with this!

mphoward (approved):
Looks good to me!

mphoward:
@mzbush adding you as a reviewer for these changes and tests

smolyay/grid.py
```diff
@@ -0,0 +1,400 @@
...
+    ``basis_set`` is a :class:`NestedBasisFunctionSet` which contains
+    one-dimensional nested points (points at each level of approximation
+    contains the points from all previous levels of approximations), their
+    corresponding basis functions, and levels, which indicates
+    levels each constituent basis function was taken from.
+    At this point, all the data extracted from ``basis_set``
+    are unidimensional.
+
+    For a given number of independent variables (``dimension``), the
+    Smolyak method of sparse sampling, creates grid points, and their
```
mzbush, line 74:
```suggestion
    Smolyak method of sparse sampling creates grid points, and their
```

smolyay/grid.py
```diff
@@ -0,0 +1,400 @@
...
+import numpy
+
+from smolyay.basis import (BasisFunctionSet, NestedBasisFunctionSet)
+
+
+class IndexGridGenerator(abc.ABC):
+    """Grid points for making a surrogate.
+
+    Depending on the approach and dimensionality
+    , a set of grid points, and their corresponding
```
mzbush, line 13:
```suggestion
    Depending on the approach and dimensionality,
    a set of grid points, and their corresponding
```

smolyay/grid.py
```diff
@@ -0,0 +1,400 @@
...
+    For instance:
+    ..math::
+        if n = 2 and \mu = 1, then:
+            if \sum_{1}^{n} k_i = 2:
+                (k_1 = 1, k_2 = 1)
+            if \sum_{i=1}^{n} k_i = 3:
+                (k_1 = 1, k_2 = 2)
+                (k_1 = 2, k_2 = 1)
+
+    An object of this class, generates the indexes of
```
mzbush, line 105:
```suggestion
    An object of this class generates the indexes of
```

smolyay/grid.py
```diff
@@ -0,0 +1,400 @@
...
+    replaced with indexes based on the points' distribution at each
+    indicy, determined by levels (each indicy is the index of the levels,
+    extracted from ``basis_set``). Then, grid points' indexes are
+    generated by making all combinations of indexes for
+    each combination of levels.
+
+    Once grid points' indexes are generated, actual grid points
+    and their corresponding basis functions can be made by simply replacing
+    the indexes with their corresponding points and basis functions.
+    multi-dimensional levels (levels' indexes each basis function was taken
```
mzbush, line 118:
```suggestion
    Multi-dimensional levels (levels' indexes each basis function was taken
```

mzbush (approved):

cakieslich (approved):

PR #4, opened 2022-11-01
Feature/surrogate1

mzf0069:
`class Surrogate ` creates a framework for approximation of a given function based on a set of grid points and basis functions.


smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+        # evaluate real function at grid points
+        real_function_at_grids = numpy.zeros((len(self._grids.points)))
+        for point_num in range(len(self._grids.points)):
+            real_function_at_grids[point_num] = (
+                self.real_function(*transformed_grids[point_num]))
+
+        # fit
+        self._coefficients = solve_linear(
+            self._basis_matrix, real_function_at_grids,
+            self.linear_solver)
```
mphoward, line 170: I don't quite understand why this can't be:
```suggestion
        # fit
        self.train_from_data(real_function_at_grids, linear_solver)
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+        list
+            Coefficients of the surrogate.
+
+        Raises
+        ------
+        IndexError
+            Data should be of length of the grid points.
+        """
+        if len(real_function_at_grids) != len(self._grids.points):
+            raise IndexError("Data must be in length of the grid points.")
```
mphoward, line 198: You should force the data to be an `ndarray` with `ndmin=2`, and then check the shape
```suggestion
        if real_function_at_grids.shape != self._grids.points.shape:
            raise IndexError("Data must be in length of the grid points.")
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+            For surrogate to be evaluated, function needs to be trained.
+
+        IndexError
+            Input of the surrogate must be of length of the
+            real function's dimensionality.
+        """
+        if self._coefficients is None:
+            raise ValueError('Function needs training!')
+        if isinstance(x, (int, float)):
+            x = [x]
```
mphoward, line 233: I think:
```suggestion
        x = numpy.atleast_1d(x)
```
would work.

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+        if len(x) != self._dimension:
```
mphoward, line 234: If you cast to ndarray, this would be a stricter test:
```suggestion
        if x.shape != (self._dimension,):
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+            for dimension_ in range(self._dimension):
+                output *= (self._grids.basis_functions[index_point]
+                           [dimension_](input_surrogate[dimension_]))
+            output *= self._coefficients[index_point]
+            surrogate_output += output
+
+        return surrogate_output
+
+
+def solve_linear(a, b, method):
```
mphoward, line 257: If you can make `train()` use `train_from_data()`, then these if statements could just be put in `train_from_data` rather than as a separate function.

mphoward (changes requested):
This is a good start! Here are some initial suggestions.

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+                                         len(self._grids.points)))
+        point_num = 0
+        for points in self._grids.points:
+            for dimension_ in range(self._dimension):
+                grid_num = 0
+                for basis_function in list(zip(
+                        *self._grids.basis_functions))[dimension_]:
+                    self._basis_matrix[point_num][grid_num] *= (
+                        basis_function(points[dimension_]))
+                    grid_num += 1
```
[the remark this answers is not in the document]
mzf0069: Yes. This is evaluating products of basis function at grid points.

smolyay/surrogate.py
```diff
@@ -0,0 +1,279 @@
...
+    @grids_generator.setter
+    def grids_generator(self, grids_generator):
+        if not isinstance(grids_generator, IndexGridGenerator):
+            raise TypeError('Grids must be IndexGridGenerator')
+        self._grids_generator = grids_generator
+
+    def _make_basis_matrix(self):
+        """Generate basis matrix."""
+        self._basis_matrix = numpy.ones((len(self._grids.points),
+                                         len(self._grids.points)))
```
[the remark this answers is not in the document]
mzf0069: This wouldn't work because self._grid is an object containing the points and functions.

smolyay/surrogate.py
mzf0069: This wouldn't work because shape of the `grid_points` is (# of grid points, dimension). However, shape of the `real_function_at_grids` is (1, # of grid points). 
This however work: 
`numpy.array(real_function_at_grids, ndmin=2).shape[1] != (numpy.array(self.grid.points).shape[0])`

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+            Coefficients of the surrogate.
+        """
+        # transform grid points
+        transformed_grids = numpy.zeros((
+            numpy.array(self.grid.points).shape[0], self._domain.shape[0]))
+        for dimension in range(self._domain.shape[0]):
+            transformed_grids[:, dimension] = (
+                numpy.polynomial.polyutils.mapdomain(
+                    numpy.array(self.grid.points)[:, dimension],
+                    (-1, 1), self.domain[dimension]))
```
mphoward, line 174: This is clever that there is a NumPy tool for doing this, but it is a little limiting and unfortunate that it only works one dimension at a time. The equations from their documentation (https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polyutils.mapdomain.html) are easy to generalize to an `(N,dim)` array though:

```
def _mapdomain(self, x, old, new):
    old = numpy.array(old, copy=False, ndmin=2)
    new = numpy.array(new, copy=False, ndmin=2)

    # error checking
    if old.shape != new.shape:
        raise TypeError('Old and new domain must have the same shape')
    if old.shape[1] != 2:
        raise TypeError('Domain should be a dim x 2 array')
    if old.shape[0] != x.shape[1]:
        raise TypeError('Points must be an N x dim array matching domain')

    new_x = new[:,0] + (new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
    if old.shape[0] == 1:
        new_x = new_x[0]
    return new_x
```
This can be run in either direction
```
real_points = self._mapdomain(scaled_points, np.array([[-1, 1]]*self.dimension), self.domain)
scaled_points = self._mapdomain(real_points, self.domain, np.array([[-1, 1]]*self.dimension))
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # evaluate real function at grid points
+        real_function_at_grids = numpy.zeros((numpy.array(
+            self.grid.points).shape[0]))
+        for point_num in range(numpy.array(self.grid.points).shape[0]):
+            real_function_at_grids[point_num] = (
+                real_function(*transformed_grids[point_num]))
```
mphoward, line 181: You can do this with list comprehension:
```suggestion
        data = [function(point) for point in self.points]
```
In writing this, I am suggesting that you expose a `points` property of the surrogate that gives the points where sampling should be done:
```
@property
def points(self):
    if self._points is None:
        self._points = self._mapdomain(self.grid.points, [[-1, 1]]*self.dimension, self.domain)
    return self._points
```
Make sure to add `self._points = None` to the constructor and the `_reset_grid()` method.

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+                self._make_basis_matrix(),
+                real_function_at_grids).tolist()
+        elif linear_solver == 'inv':
+            self._coefficients = numpy.dot(numpy.linalg.inv(
+                self._make_basis_matrix()),
+                real_function_at_grids).tolist()
+        elif linear_solver == 'lstsq':
+            self._coefficients = numpy.linalg.lstsq(self._make_basis_matrix(),
+                                                    real_function_at_grids,
+                                                    rcond=None)[0].tolist()
```
mphoward, line 225: Make sure to catch unknown solvers, and reduce duplication:
```suggestion
        basis_matrix = self._make_basis_matrix()
        if linear_solver == 'lu':
            self._coefficients = numpy.linalg.solve(basis_matrix, data)
        elif linear_solver == 'inv':
            self._coefficients = numpy.dot(numpy.linalg.inv(basis_matrix), data)
        elif linear_solver == 'lstsq':
            self._coefficients = numpy.linalg.lstsq(basis_matrix, data, rcond=None)[0]
        else:
            raise ValueError('Solver not recognized')
```
You can call the `tolist()` method in the coefficients getter if you so desire, no need to use it here since it will probably be convenient to keep this as a numpy array.

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+            For surrogate to be evaluated, function needs to be trained.
+        IndexError
+            Input of the surrogate must be of length of the
+            real function's dimensionality.
+        """
+        if self._coefficients is None:
+            raise ValueError('Function needs training!')
+        x = numpy.atleast_1d(x)
+        if x.shape != (self._domain.shape[0],):
+            raise IndexError('Input must be of length of the dimension.')
```
mphoward, line 254:
```suggestion
        x = numpy.array(x, copy=False, ndmin=1)
        if x.shape != (self.dimension, ):
            raise IndexError('Input must match dimension of domain.')
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # transform grids into basis domain
+        input_surrogate = []
+        for dimension in range(self._domain.shape[0]):
+            basis_domain_grid = numpy.polynomial.polyutils.mapdomain(
+                x[dimension], self.domain[dimension], (-1, 1))
+            input_surrogate.append(basis_domain_grid)
```
mphoward, line 261:
```suggestion
        x_ = self._mapdomain(x, self.domain, [[-1,1]]*self.dimension)
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+        # evaluate surrogate
+        output = numpy.ones((len(self.grid.basis_functions)))
+        for i, input_ in enumerate(input_surrogate):
+            for j in range(len(self.grid.basis_functions)):
+                output[j] *= list(zip(
+                    *self.grid.basis_functions))[i][j](input_)
+        surrogate_output = numpy.sum(
+            numpy.multiply(self._coefficients, output))
+
+        return surrogate_output
```
mphoward, line 272:
```suggestion
        return value
```

mphoward (changes requested):

smolyay/surrogate.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # evaluate real function at grid points
+        real_function_at_grids = numpy.zeros((numpy.array(
+            self.grid.points).shape[0]))
+        for point_num in range(numpy.array(self.grid.points).shape[0]):
+            real_function_at_grids[point_num] = (
+                real_function(*transformed_grids[point_num]))
+
+        # fit
+        return self.train_from_data(real_function_at_grids, linear_solver)
```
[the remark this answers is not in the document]
mzf0069: Should I change `real_function_at_grids` to `data`? or `function_at_grids` would be a better option since `Surrogate.train` trains a surrogate of a callable?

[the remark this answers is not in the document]
mphoward: I would call it `data`. `train` is just a pass through function: it creates the `data` then calls `train_from_data`

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+            return self._domain[0].tolist()
+
+    @domain.setter
+    def domain(self, value):
+        self._domain = numpy.array(value, ndmin=2)
+        self._reset_grid()
+
+    @property
+    def grid(self):
+        """:class:`IndexGrid` data structure."""
```
mphoward, line 124:
```suggestion
        """:class:`IndexGrid`: Sampling grid."""
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+        if self._grid is None:
+            self._grid = self.grid_generator(self.dimension)
+        return self._grid
+
+    @property
+    def grid_generator(self):
+        """:class:`IndexGridGenerator`: Generator used for making the grids."""
```
mphoward, line 131:
```suggestion
        """:class:`IndexGridGenerator`: Sampling grid generator."""
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Raises
+        ------
+        IndexError
+            Data should be of length of the grid points.
+        ValueError
+            Solver should be selected between defined methods.
+        """
+        data = numpy.array(data, copy=False, ndmin=1)
+        if data.shape != (self.points.shape[0],):
```
mphoward, line 226: Using `grid.points` here will allow us to bypass creating `points` if it is not actually needed:
```suggestion
        if data.shape != (self.grid.points.shape[0],):
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+            raise IndexError("Data must be same length as grid points.")
+        basis_matrix = self._make_basis_matrix()
```
mphoward, line 228: Since `basis_matrix` is only used in this function, I think it would make more sense to just put that code here (it is only a few lines)

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+                basis_matrix),
+                data)
+        elif linear_solver == 'lstsq':
+            self._coefficients = numpy.linalg.lstsq(basis_matrix,
+                                                    data,
+                                                    rcond=None)[0]
+        else:
+            raise ValueError('Solver not recognized')
+
+    def _make_basis_matrix(self):
```
mphoward, line 244: See comment above to fuse this code with above and remove this function.

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+        for i, point in enumerate(points):
+            for j, basis in enumerate(basis_functions):
+                if self.dimension > 1:
+                    value = numpy.prod([f(x) for x, f in zip(point, basis)])
+                else:
+                    value = basis(point)
+                basis_matrix[i, j] = value
+        return basis_matrix
+
+    def _mapdomain(self, x, old, new):
```
mphoward, line 270: The only part where this actually uses `self` is to check the dimensionality, so let's remove that and make this a static method:
```suggestion
    @staticmethod
    def _mapdomain(x, old, new):
```

smolyay/surrogate.py
```diff
@@ -0,0 +1,309 @@
...
+            Domain should be a dim x 2 array.
+            Points must be an N x dim array matching domain.
+        """
+        old = numpy.array(old, copy=False, ndmin=2)
+        new = numpy.array(new, copy=False, ndmin=2)
+
+        # error checking
+        if old.shape != new.shape:
+            raise TypeError('Old and new domain must have the same shape')
+        if old.shape != (self.dimension, 2):
```
mphoward, line 300:
```suggestion
        if old.ndim != 2 or old.shape[1] != 2:
```

tests/grid_test.py
```diff
@@ -22,125 +21,56 @@ def test_error_nested_basis_set_type():
...
-            test_grid_points_basis_eval_two_dimension[grid_index].append(
-                grid_basis_function.__call__(test_point))
-    expected_grid_points_basis_two_dimension = [[1, 1], [0.72, 1],
-                                                [2*(0.72**2)-1, 1],
-                                                [1, 0.72], [1, 2*(0.72**2)-1]]
+    grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(2))
+    grid = grid_gen(1)
+    assert grid.indexes == [0, 1, 2, 3, 4]
+    assert grid.levels == [0, 1, 1, 2, 2]
+    assert numpy.allclose(grid.points, [0, -1, 1, -0.70710, 0.70710], atol=1e-10)
```
mphoward, line 30: Is there a reason to make atol stricter than the default? Suggest to remove:
```suggestion
    assert numpy.allclose(grid.points, [0, -1, 1, -0.70710, 0.70710])
```

tests/grid_test.py
```diff
@@ -22,125 +21,56 @@ def test_error_nested_basis_set_type():
...
+                           [1, 1], [0, 2], [0, 2]]
+    assert numpy.allclose(grid.points,
+                          [[0, 0], [-1, 0], [1, 0], [0, -1],
+                           [0, 1],
+                           [-0.70710, 0], [0.70710, 0],
+                           [-1, -1],
+                           [-1, 1], [1, -1], [1, 1],
+                           [0, -0.70710],
+                           [0, 0.70710]],
                           atol=1e-10)
```
mphoward, line 52:
```suggestion
                           [0, 0.70710]])
```

tests/grid_test.py
```diff
@@ -22,125 +21,56 @@ def test_error_nested_basis_set_type():
...
 
+def test_tensor_grid_1d():
+    grid_gen = TensorGridGenerator(ChebyshevFirstKind.make_nested_set(2))
+    grid = grid_gen(1)
+    assert grid.indexes == [0, 1, 2, 3, 4]
+    assert numpy.allclose(grid.points, [0, -1, 1, -0.70710, 0.70710], atol=1e-10)
```
mphoward, line 58:
```suggestion
    assert numpy.allclose(grid.points, [0, -1, 1, -0.70710, 0.70710])
```

tests/grid_test.py
```diff
@@ -22,125 +21,56 @@ def test_error_nested_basis_set_type():
...
+    assert grid.indexes == [[0, 0], [0, 1], [0, 2],
+                            [1, 0],
+                            [1, 1], [1, 2], [2, 0],
+                            [2, 1], [2, 2]]
+    assert numpy.allclose(
+        grid.points,
+        [[0, 0], [0, -1], [0, 1], [-1, 0],
+         [-1, -1], [-1, 1], [1, 0],
+         [1, -1], [1, 1]],
+        atol=1e-10)
```
mphoward, line 73:
```suggestion
         [1, -1], [1, 1]])
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    x1, x2 = x
+    branin1 = (x2 - 5.1 * x1 ** (2)/(4 * numpy.pi ** 2)
+               + 5 * x1 / (numpy.pi) - 6) ** 2
+    branin2 = 10 * (1 - 1 / (8 * numpy.pi)) * numpy.cos(x1)
+    branin3 = 10
+    branin_function = branin1 + branin2 + branin3
+    return branin_function
+
+
+def branin_at_grids():
```
mphoward, line 43: This should just be done in the test using that grid's points, remove this.

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+        ChebyshevFirstKind.make_nested_set(2))
+    domain = (-1, 1)
+    surrogate_object = Surrogate(domain, grid_generator)
+    assert numpy.allclose(surrogate_object.domain,
+                          numpy.array(domain, ndmin=2))
+    assert surrogate_object._coefficients is None
+    assert surrogate_object._grid is None
+    grid_object = surrogate_object.grid
+    assert isinstance(grid_object, IndexGrid)
+    assert surrogate_object.dimension == 1
```
mphoward, line 72: Test as it should be used:
```suggestion
    assert numpy.allclose(surrogate_object.domain, domain)
    assert surrogate_object.coefficients is None
    assert isinstance(surrogate_object.grid, IndexGrid)
    assert surrogate_object.dimension == 1
```
Should you also check `points` here?

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def test_initialization_2d():
+    """Test if class is properly intiallized."""
+    grid_generator = SmolyakGridGenerator(
+        ChebyshevFirstKind.make_nested_set(2))
+    domain = [(-5, 5), (-8, 0)]
+    grid_generator_ = 'not an IndexGridGenerator'
```
mphoward, line 80: Try to make your tests easier to read: add white space, and on't add variables until you need them:
```suggestion
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    surrogate_object = Surrogate(domain, grid_generator)
+    assert numpy.allclose(surrogate_object.domain,
+                          numpy.array(domain, ndmin=2))
+    with pytest.raises(TypeError):
+        Surrogate(domain, grid_generator_)
+    assert surrogate_object._coefficients is None
+    grid_object = surrogate_object.grid
+    assert isinstance(grid_object, IndexGrid)
+    assert surrogate_object.dimension == 2
```
mphoward, line 89:
```suggestion
    assert numpy.allclose(surrogate_object.domain, domain)
    assert surrogate_object.coefficients is None
    assert isinstance(surrogate_object.grid, IndexGrid)
    assert surrogate_object.dimension == 2
```

This should go in its own test, it does something different:
```
    with pytest.raises(TypeError):
        Surrogate(domain, 'not a grid generator')
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    """Test if a correct basis matrix is generated."""
+    grid_generator = SmolyakGridGenerator(
+        ChebyshevFirstKind.make_nested_set(1))
+    surrogate_object = Surrogate([(-7, 3), (-18, 9)], grid_generator)
+    expected_basis_matrix = numpy.array(
+        [[1, 0, -1, 0, -1], [1, -1, 1, 0, -1], [1, 1, 1, 0, -1],
+         [1, 0, -1, -1, 1], [1, 0, -1, 1, 1]])
+    assert numpy.allclose(expected_basis_matrix,
+                          surrogate_object._make_basis_matrix())
+
```
mphoward, line 124: This is an implementation detail that the object is free to change, so don't test it like this. Instead, make sure that training happens correctly elsewhere.

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    surrogate_object = Surrogate([(-1, 1), (-1, 1), (-1, 1)],
+                                 SmolyakGridGenerator(
+                                     ChebyshevFirstKind.make_nested_set(2)))
+    surrogate_object.train(function_0, 'lu')
+    test_point = (0.649, 0, -0.9)
+    test_point_ = (-0.885, 1, 0.275)
+    assert numpy.allclose(surrogate_object(test_point),
+                          function_0([0.649, 0, -0.9]))
+    assert numpy.allclose(surrogate_object(test_point_),
+                          function_0([-0.885, 1, 0.275]))
```
mphoward, line 137: This would be much more readable, try applying this to your other tests:
```suggestion
    grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(2))
    surrogate = Surrogate([(-1, 1), (-1, 1), (-1, 1)], grid_gen)
    surrogate.train(function_0, 'lu')

    points = [(0.649, 0, -0.9), (-0.885, 1, 0.275)]
    surrogate_values = [surrogate(x) for x in points]
    exact_values = [function_0 for x in points]
    assert numpy.allclose(surrogate_values, exact_values)
```
Also, is there a reason that you chose these specific points? They seem weird, so a comment would be helpful.

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+def test_surrogate_0_shifted():
+    """Test if surrogate generates exact results for a shifted function 0."""
+    surrogate_object = Surrogate([(-1, 1), (-1, 1), (-1, 1)],
+                                 SmolyakGridGenerator(
+                                     ChebyshevFirstKind.make_nested_set(2)))
+    surrogate_object_ = Surrogate([(0, 2), (-2, 0), (-1, 1)],
+                                  SmolyakGridGenerator(
+                                      ChebyshevFirstKind.make_nested_set(2)))
+    surrogate_object.train(function_0, 'lu')
+    surrogate_object_.train(function_0_shifted, 'lu')
```
mphoward, line 162: I didn't notice before, but this is a case where it would be helpful to have saved the training data:
```suggestion
    surrogate_object.train(function_0, 'lu')
    surrogate_object_.train_from_data(surrogate_object.data, 'lu')
```
You should go add this is as a property of the surrogate, and set its value in `train_from_data`.

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    assert numpy.allclose(surrogate_object(test_point_),
+                          function_1([-0.885, 1]))
+
+
+def test_surrogate_0_shifted():
+    """Test if surrogate generates exact results for a shifted function 0."""
+    surrogate_object = Surrogate([(-1, 1), (-1, 1), (-1, 1)],
+                                 SmolyakGridGenerator(
+                                     ChebyshevFirstKind.make_nested_set(2)))
+    surrogate_object_ = Surrogate([(0, 2), (-2, 0), (-1, 1)],
```
mphoward, line 158: You do use the trailing `_` when you want to use a name that you've already used elsewhere, but it's confusing in this context where you're comparing two models. Why not use `surrogate_1` and `surrogate_2`?

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+    """Test if train_from_data_works."""
+    surrogate_object = Surrogate([(-5, 10), (0, 15)], SmolyakGridGenerator(
+        ChebyshevFirstKind.make_nested_set(4)))
+    surrogate_object_ = Surrogate([(-5, 10), (0, 15)], SmolyakGridGenerator(
+        ChebyshevFirstKind.make_nested_set(4)))
+    surrogate_object.train_from_data(branin_at_grids(), 'lu')
+    assert numpy.allclose(surrogate_object((8, 0.75)),
+                          branin([8, 0.75]), atol=1e-10)
+    with pytest.raises(IndexError):
+        surrogate_object_.train_from_data(branin_at_grids()[:5], 'lu')
```
mphoward, line 187: Don't make two objects when 1 would suffice, and also test how you want it to be used:
```suggestion
    data = branin(surrogate_object.points)
    surrogate_object.train_from_data(data)
    assert surrogate_object((8, 0.75)) == pytest.approx(branin([8, 0.75]))

    with pytest.raises(IndexError):
        surrogate_object.train_from_data(data[:5])
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,206 @@
...
+
```
mphoward, line 188: Should you add tests for the different solver methods? You could also parametrize the tests on this, so all will run with all solvers?

mphoward (changes requested):
This is getting close!

tests/surrogate_test.py
mzf0069: I just picked some random points from the domain.

tests/surrogate_test.py
mzf0069: for some reason `pytest.approx` doesn't work. It gives this error: 
`      assert 9.219799831341572 == 9.219766083043577 ± 9.2e-06
         comparison failed
        Obtained: 9.219799831341572
         Expected: 9.219766083043577 ± 9.2e-06
`
However, this works: 
`numpy.allclose(surrogate(point), branin([8, 0.75]))`

mphoward: Per the error message, the two numbers do not agree within the required relative tolerance:
```
>>> 9.219799831341572-9.219766083043577
3.374829799440704e-05
```
`numpy.allclose` uses a different tolerance. If you are sure the agreement is good enough, you can do one of two things:
* Relax the tolerance for `pytest.approx`, or
* Use `numpy.isclose` (which is the scalar version of `allclose`)

tests/surrogate_test.py
mphoward: Yes it can. You called `surrogate_object.train` in line 161. If you store its `data` in a property as suggested, it will be accessible when you call `surrogate_object_.train_from_data` (which is a different surrogate) on line 162.

tests/surrogate_test.py
```diff
@@ -0,0 +1,201 @@
...
+                          surrogate_2. coefficients)
+
+
+# evaluate function_2 at test point for parametrization
+f_2_point = function_2(0.367)
+
+
+@pytest.mark.parametrize('linear_solver, exact_value',
+                         [('lu', f_2_point), ('lstsq', f_2_point),
+                          ('inv', f_2_point)])
```
mphoward, line 151:
```suggestion
@pytest.mark.parametrize('linear_solver', ['lu','lstsq','inv'])
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,201 @@
...
+def test_surrogate_2(linear_solver, exact_value):
+    """Test if surrogate generates exact results for simple 1D function."""
+    grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(2))
+    surrogate = Surrogate((-10, 10), grid_gen)
+    surrogate.train(function_2, linear_solver)
+    # random point in the domain
+    point = 0.367
+    assert numpy.isclose(surrogate(point), exact_value)
```
mphoward, line 159:
```suggestion
    assert numpy.isclose(surrogate(point), function_2(point))
```

tests/surrogate_test.py
```diff
@@ -0,0 +1,201 @@
...
+    surrogate = Surrogate([(-1, 1), (-1, 1)], grid_gen)
+    surrogate.train(function_1, 'lstsq')
+    # random points in the domain
+    points = [(0.649, -0.9), (-0.885, 1)]
+    surrogate_values = [surrogate(x) for x in points]
+    exact_values = [function_1(x) for x in points]
+    assert numpy.allclose(surrogate_values, exact_values)
+
+
+def test_surrogate_0_shifted():
```
mphoward, line 134: I see how this could work for testing a shifted function, but I don't think it quite does enough. It would be good if you added a step where you evaluated both surrogates at effectively the same point (with shifting for `surrogate_2`) to make sure it actually evaluates the same.

smolyay/surrogate.py
```diff
@@ -0,0 +1,311 @@
...
+            For surrogate to be evaluated, function needs to be trained.
+        IndexError
+            Input of the surrogate must be of length of the
+            function's dimensionality.
+        """
+        if self._coefficients is None:
+            raise ValueError('Function needs training!')
+        x = numpy.array(x, copy=False, ndmin=1)
+        if x.shape != (self.dimension, ):
+            raise IndexError('Input must match dimension of domain.')
```
mphoward, line 188: You should probably add an error check here that x lies within the domain:
```
if not (numpy.all(x >= self._domain[:,0]) and numpy.all(x <= self._domain[:,1])):
    raise ValueError('x must lie in domain of surrogate')
```
Then add a test for this too. I think we might have discussed how the basis functions _should_ catch this themselves, but having the error raised here will make more sense to the caller.

mphoward (approved):
Just a few more small changes from my side: this is good to go once you make them! Approving now assuming you implement them all.

mphoward (approved):

cakieslich (approved):

PR #6, opened 2023-01-16
Fix for slow exponential growth

mzbush:
Creating grid points for slow exponential growth requires that the levels property in NestedBasisFunctionSet can sometimes have empty elements.


tests/slow_growth_test.py
```diff
@@ -0,0 +1,27 @@
+import pytest
```
mphoward, line 1: Please add these tests to the existing file where the other tests for `SmolyakGridGenerator` are, rather than a separate file.

mphoward (approved):
This PR can be merged once the tests are moved to the appropriate file. Thanks Michaela!

tests/slow_growth_test.py
mzbush: Done.

mphoward:
This is ready to be merged.

cakieslich (approved):

PR #7, opened 2023-01-25
Feature/test functions

mzbush:
Add class for testing functions


smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
+import abc
```
mphoward, line 1: Please rename this file `benchmark.py` so that it can be imported like `smolyay.benchmark`.

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+
+import numpy
+
+class test_fun(abc.ABC):
```
mphoward, line 5: Please rename this class either `BenchmarkProblem` or `BenchmarkFunction`. Per PEP8, this is the appropriate naming convention for classes. We are using non-PEP8 names for the implementations below because these are accepted elsewhere.

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+
+    lower_bounds : list of floats of size 1xdim
+        the lower bounds of the domain of each variable
+
+    upper_bounds : list of floats of size 1xdim
+        the upper bounds of the domain of each variable
+
+    domain : list of lists
+        the domain of the inputs for each dimension
+    """
```
mphoward, line 28: Keep docstrings current, descriptive, and correct. Docstrings should describe what the object does, not its     implementation details.
```suggestion
    """Benchmark function.

    <longer description of what a benchmark problem is>
    """
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+
+        Parameters
+        ----------
+        x : list of floats
+            the input to the call method
+
+        Raises
+        ------
+        ValueError
+            if the input is outside is outside the domain
```
mphoward, line 40: Docstrings should be meaningful of what the parameters represent, not things like "input to the method" (that is obvious because it is an argument).
```suggestion
        """Evaluate the function.

        Parameters
        ----------
        x : list
            Function input.

        Raises
        ------
        ValueError
            If the input is outside the function domain.
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+        """
+        if self.dim > 1:
+            oob = any(
+                xi < bound[0]-1e-10 or xi > bound[1]+1e-10 for xi,bound in
```
mphoward, line 44: The domain is closed, so you can compare using < and >. If you are getting an error elsewhere, that is an issue with how that value of x is being generated.
```suggestion
                xi < bound[0] or xi > bound[1] for xi,bound in
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+                zip(x,self.domain))
+        else:
+            oob = x > self.upper_bounds[0] or x < self.lower_bounds[0]
```
mphoward, line 47: This is better because it accesses the data directly, rather than an indirect access that constructs a list. It also uses the same pattern as above, so it is less confusing to read.
```suggestion
            oob = x < self.domain[0][0] or x > self.domain[0][1]
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+        if oob:
+            raise ValueError("Input out domain of function.")
+        return self._function(x)
+    
+    @property
+    def name(self):
+        """Name of the function"""
+        return self.__class__.__name__
```
mphoward, line 55: Suggest to use this because it is easier to read (avoids one dunder, which can be confusing)
```suggestion
        return type(self).__name__
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+    @property
+    def dim(self):
+        """Number of variables"""
+        return len(self.domain)
+
+    @property
+    def lower_bounds(self):
+        """ the lower bounds of the domain of each variable"""
+        lower, _ = zip(*self.domain)
+        return list(lower)
```
mphoward, line 66: List comprehension is more direct here. Docstrings should for properties should also look like this:
```suggestion
        """list: Lower bound of each variable."""
        return [bound[0] for bound in self.domain]
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+        return self._function(x)
+    
+    @property
+    def name(self):
+        """Name of the function"""
+        return self.__class__.__name__
+
+    @property
+    def dim(self):
+        """Number of variables"""
```
mphoward, line 59: For consistency with `Surrogate`:
```suggestion
    def dimension(self):
        """int: Number of variables."""
```

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+        return len(self.domain)
+
+    @property
+    def lower_bounds(self):
+        """ the lower bounds of the domain of each variable"""
+        lower, _ = zip(*self.domain)
+        return list(lower)
+    
+    @property
+    def upper_bounds(self):
```
mphoward, line 69: Apply same changes as for `lower_bounds` here.

smolyay/test_function_class.py
```diff
@@ -0,0 +1,6478 @@
...
+        """the upper bounds of the domain of each variable"""
+        _, upper = zip(*self.domain)
+        return list(upper)
+
+    @property
+    @abc.abstractmethod
+    def domain(self):
+        """the lower and upper bounds of each variable"""
```
mphoward, line 77:
```suggestion
        """list: Domain of the function.
        
        The domain must be specified as lower and upper bounds for each variable as a list of lists.
        """
```

mphoward (changes requested):
Thanks Michaela! I commented only on the base implementation of this because the file is very large to review. Given that copy-pasting seemed rather successful for the other functions, maybe you can do that for all of them so that we are confident the implementations come from the same source code? (And not have to check ones that you implemented in a different way)

We still need to check what the domains should be with @cakieslich

smolyay/test_function_class.py
mzbush: The original reason that I put that there was because there were rare occasions in which while trying to get the samples for training the surrogate model, the function would give the out of bounds error. The issue appears to be rounding from floating point arithmetic in the _mapdomain function in the surrogate class.

For example, trying to make a surrogate function for the BenchmarkFunction ex8_1_3, the surrogate wants to sample from the point (8.460000000000003, -0.8800000000000008) but the domain of the first variable is [-10.6, 8.46].

mphoward: I see. We should then take a look at that method and see if we can improve it in some way. One option would be to clamp the values to lie within the domain.

mzbush: I tried to change the _mapdomain function to see if I could force the bounds using numpy.interp. However, the solution I came up with that passes all the pytests is very messy.

It replaces the original return statement with this:
`x = numpy.array(x, copy=False, ndmin=1)`
`new_x = numpy.zeros(x.shape)`
`if not 1 in old.shape:`
`    # we have more than one domain`
`    if not len(x.shape) == 1:`
`        # many values in x share a domain`
`        for i in range(x.shape[-1]):`
`            new_x[:,i] = numpy.interp(x[:,i],old[i,:],new[i,:])`
`    else:`
`        # each value in x use different domains`
`        for i in range(x.shape[-1]):`
`            new_x[i] = numpy.interp(x[i],old[i,:],new[i,:])`
`else:`
`    # all values in x use the same domain`
`    new_x = numpy.interp(x,old[0],new[0])`
`return new_x`

numpy.interp keeps everything in the bounds just fine, but it's very picky about the shape of old and new, and it doesn't appear to want to do multiple domains at once.

mphoward: How about something simpler like this?
```
# clamp lower bound
flags = new_x < new[:, 0]
new_x[flags] = new[flags, 0]

# clamp upper bound
flags = new_x > new[:, 1]
new_x[flags] = new[flags, 1]
```

mzbush: Yeah that's much simpler.

smolyay/benchmark/benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            oob = (numpy.any(x < self.domain[0][0]) or 
+                    numpy.any(x > self.domain[0][1]))
+        if oob:
+            raise ValueError("Input out domain of function.")
+        return numpy.squeeze(self._function(x))
+    
+    @property
+    def name(self):
+        """Name of the function"""
```
mphoward, line 43:
```suggestion
        """str: Name of the function"""
```

smolyay/benchmark/benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+        ValueError
+            If the input is outside the function domain.
+        """
+        x = numpy.array(x, copy=False,ndmin=2)
+        if self.dimension > 1:
+            oob = any(numpy.any(xi < bound[0]) or numpy.any(xi > bound[1]) 
+                    for xi,bound in zip(x.transpose(),self.domain))
+        else:
+            oob = (numpy.any(x < self.domain[0][0]) or 
+                    numpy.any(x > self.domain[0][1]))
```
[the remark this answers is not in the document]
mzbush: It actually is because for you're passing multiple inputs for a single dimension, x[...,0] will only pass the first input, not all of the inputs.

smolyay/surrogate.py
```diff
@@ -361,7 +361,16 @@ def _mapdomain(x, old, new):
...
-        return new[:,0]+(new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
+        new_x = new[:,0]+(new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
+        # clamp lower bound
+        flags = numpy.array(new_x < new[:, 0])
+        test = numpy.broadcast_to(new[:, 0],flags.shape)
+        new_x[flags] = test[flags]
+        # clamp upper bound
+        flags = numpy.array(new_x > new[:, 1])
+        test = numpy.broadcast_to(new[:, 1],flags.shape)
+        new_x[flags] = test[flags]
```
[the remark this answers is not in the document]
mzbush: No, it gives a lot of index errors.

tests/benchmark_test.py
```diff
@@ -0,0 +1,208 @@
...
+
+@pytest.mark.parametrize("fun",functions)
+@pytest.mark.filterwarnings("error")
+def test_functions_call(fun):
+    """Test all functions do not raise error or warning for inputs in bounds"""
+    x_list = numpy.linspace(fun.lower_bounds,fun.upper_bounds)
+    y_call = [fun(i) for i in x_list]
+    y_call_multi = fun(x_list)
+    y_fun = numpy.squeeze([fun._function(i) for i in x_list])
+    y_fun_multi = fun._function(x_list)
```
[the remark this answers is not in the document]
mzbush: It was to ensure that going through call and checking the bounds didn't mess up the structure of the inputs in a way that causes `__call__` and `_function` to behave differently.

smolyay/surrogate.py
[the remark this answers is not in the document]
mphoward: What is the test case that fails? I just ran this on my desktop and it worked fine:
```
>>> x = [2.1, -6]
>>> old = np.array([[0, 2],[-5, 5]])
>>> new = np.array([[-1, 1], [-1, 1]])
>>> new_x = new[:,0]+(new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
>>> new_x
array([ 1.1, -1.2])
>>> flags = new_x < new[:, 0]
>>> new_x[flags] = new[flags, 0]
>>> flags = new_x > new[:, 1]
>>> new_x[flags] = new[flags, 1]
>>> new_x
array([ 1., -1.])
```

tests/benchmark_test.py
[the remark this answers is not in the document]
mphoward: Couldn't you check that `__call__` returns an output with the expected shape? The cases I would check are
* If `x` is a single point, a scalar is returned.
* If `x` is 2d, an array is returned.
* If `x` is anything higher dimensional, the return value is a multidimensional array with shape `x.shape[:-1]`.

smolyay/surrogate.py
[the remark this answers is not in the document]
mzbush: When running the pytest, I got 28 errors.
FAILED tests/surrogate_test.py::test_initialization_1d - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_gradient_initialization_1d - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_initialization_2d - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_gradient_initialization_2d - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_map_function_1d - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_map_function_2d - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_0[lu] - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_0[lstsq] - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_0[inv] - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_1 - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_0_shifted - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_surrogate_2[lu] - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_surrogate_2[lstsq] - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_surrogate_2[inv] - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_train_from_data - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_gradient_train_from_data - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_gradient_surrogate_3 - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...
FAILED tests/surrogate_test.py::test_gradient_surrogate_4 - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_error_solver - IndexError: boolean index did not match indexed array along dimension 0; di...
FAILED tests/surrogate_test.py::test_error_input_surrogate - IndexError: too many indices for array: array is 2-dimensional, but 3 were ...

smolyay/benchmark/benchmark.py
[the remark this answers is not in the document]
mphoward: Oh, I think you are talking about a specific use case where the dimension of the function is 1, and you are allowing a 1d array to be interpreted as multiple inputs? This array structure is different from how the inputs are expected for higher dimensional functions. When cast to a 2d array, it has the wrong dimensionality in the second index:
```
>>> x = np.array([1, 2, 3], ndmin=2)
>>> x
array([[1, 2, 3]])
>>> x.shape
(1, 3)
```
Whereas we would expect it to have the transposed shape if we explicitly interpreted this as an array of 1d inputs:
```
>>> y = np.array([[1],[2],[3]], ndmin=2)
>>> y
array([[1],
       [2],
       [3]])
>>> y.shape
(3, 1)
```
You could check for this special case (`self.dimension == 1 and x.shape[0]  > 1`), and transpose x before the function proceeds.

Otherwise, it is very unclear why you would do it this way.

smolyay/surrogate.py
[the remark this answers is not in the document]
mphoward: These are test failure messages, but it's impossible for me to figure out what is going wrong by you pasting a portion of the output when I also can't see your code.

You should work on a specific case that is failing, and you can ask for help on Slack (with the full error output) if you need it.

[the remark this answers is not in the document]
mzbush: The crux of the issue is that the shape of new_x and flags is the number of points you have x dimension while the shape of new is dimension x 2. In order to map booleans in flags onto new, you need to first broadcast new to the shape of new_x and flags. That's what the test variable does (which admittedly needs a better name).

Yours didn't error because all the shapes just happened to work out. new_x is (2, ), flags is (2, ), new is (2, 2), and new[:,0] is (2, ).

But in the pytests like in test_initialization_1d, new_x is (5, ), flags is (5, ), new is (1, 2), and new[:,0] is (1, ).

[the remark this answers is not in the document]
mphoward: Oh, yes, I see! In that case, I would recommend that you do a `for` loop over the dimensions (if there is more than one dimension). You don't want to use `broadcast_to` because it will do an allocation & copy every time this function gets called. There is also actually a function `clip` that would help with this. I think something like:
```
for i in range(self.dimension):
     numpy.clip(new_x[:, i], new[i, 0], new[i, 1], out=new_x[:, i])
```
You probably need to treat the 1d case specially since currently `x` is not required to be 2d.

smolyay/benchmark/benchmark.py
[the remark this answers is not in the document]
mzbush: If the case where dimension is 1 is still treated as a special case, then the if-else is just being replaced with a different if statement, which kind of defeats the whole purpose of making the if-else statement unnecessary.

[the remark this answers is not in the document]
mphoward: This is not just the 1d case, it is more specific than that. It reduces code duplication and makes the purpose of the code clearer & more consistent, since you will sanitize `x` to have the right shape in this corner case that you are allowing, which is different from all of the other use cases. Hence, subsequent usage of `x` in the function will be more unified.

So no, it does not

> defeats the whole purpose of making the if-else statement unnecessary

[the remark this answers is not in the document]
mzbush: Okay, it is now done.

smolyay/surrogate.py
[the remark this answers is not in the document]
mzbush: It works now.

smolyay/benchmark/benchmark.py
```diff
@@ -28,19 +24,19 @@ def __call__(self,x):
...
-            oob = (numpy.any(x < self.domain[0][0]) or 
-                    numpy.any(x > self.domain[0][1]))
+        if self.dimension == 1 and x.shape[-1] > 1:
+            x = numpy.reshape(x,list(x.shape)+[1])
+
+        oob = any(numpy.any(x[..., i] < self.domain[i][0]) or 
+                  numpy.any(x[..., i] > self.domain[i][1]) 
+                  for i in range(self.dimension))
         if oob:
             raise ValueError("Input out domain of function.")
```
mphoward: This can go together now:
```suggestion
        if any(numpy.any(x[..., i] < self.domain[i][0]) or 
                  numpy.any(x[..., i] > self.domain[i][1]) 
                  for i in range(self.dimension)):
            raise ValueError("Input out domain of function.")
```

smolyay/benchmark/benchmark.py
```diff
@@ -28,19 +24,19 @@ def __call__(self,x):
...
         """
         x = numpy.array(x, copy=False,ndmin=2)
-        if self.dimension > 1:
-            oob = any(numpy.any(xi < bound[0]) or numpy.any(xi > bound[1]) 
-                    for xi,bound in zip(x.transpose(),self.domain))
-        else:
-            oob = (numpy.any(x < self.domain[0][0]) or 
-                    numpy.any(x > self.domain[0][1]))
+        if self.dimension == 1 and x.shape[-1] > 1:
+            x = numpy.reshape(x,list(x.shape)+[1])
```
mphoward, line 28: Doesn't this lead to a higher dimensional array than what we want? For example, if `x` is a `(5,)` list, then it will be cast to `(1,5)` by the array command, then you are adding a 1 so its shape will be `(1,5,1)`. But, we would want the shape to be `(5,1)`.

I would take the transpose if it is a 2d array meeting all the requirements:
```suggestion
        if self.dimension == 1 and x.ndim == 2 and x.shape[0] == 1 and x.shape[1] > 1:
            x = x.T
```
Then, I would add an error check on the shape of `x` before you allow the code to proceed. Specifically, that the last element of `shape` matches dimension:
```
if x.shape[-1] != self.dimension:
    raise TypeError(...)
```

mphoward (changes requested):
Thanks! One more thing to look at and then this can merge.

smolyay/benchmark/benchmark.py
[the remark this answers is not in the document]
mzbush: As an aside, in the surrogate pull request that allows multiple inputs for call and gradient, I used a similar bounds checking method. However, surrogate allows for the domain to be a single list, so using self.domain[i][0] wouldn't work for the 1D case in the Surrogate class.

smolyay/benchmark/benchmark.py
mzbush: It doesn't really matter if there's an extra dimension on the end because numpy.squeeze later in the code reduces any extra dimensions. Tacking on a dimension of 1 to the end lets the code be more flexible if you say have a BenchmarkFunction object that's 1D and you want to be able to pass a n-dimensional grid.

mphoward: You shouldn't rely on something happening later to fix the input you've gotten. I didn't realize this was also a potential use case (1d function takes a multidimensional array that needs an axis added). I think the code you want then is:
```
if self.dimension == 1:
    if x.ndim > 2:
        # add a new axis for the last 1d
        x = x[..., numpy.newaxis]
    elif x.shape[0] == 1 and x.shape[1]  > 1:
        # the cast to 2d puts these in wrong order, so transpose
        x = x.T
```

FYI: I looked in the documentation and transpose does not actually invoke a memory copy, which is nice! However, it is cleaner to use it sparingly.

mzbush: This doesn't account for my test case where I pass an array of size (2,3) into a 1D function, so I modified the if statements. The 1D functions really shouldn't activate the wrong number of dimensions error.

smolyay/surrogate.py
```diff
@@ -361,7 +361,14 @@ def _mapdomain(x, old, new):
         if old.ndim != 2 or old.shape[1] != 2:
             raise TypeError('Domain should be a dim x 2 array')
 
-        return new[:,0]+(new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
+        new_x = new[:,0]+(new[:,1]-new[:,0])*((x-old[:,0])/(old[:,1]-old[:,0]))
+        # clamp bounds
+        if len(new_x.shape) == 1:
+            numpy.clip(new_x, new[:,0], new[:,1],out=new_x)
+        else:
+            for i in range(len(new)):
```
mphoward, line 369:
```suggestion
            for i in range(new.shape[0]):
```

mphoward (approved):
Thanks Michaela! There was one small change I wanted to make, but I will commit it and merge this.

PR #8, opened 2023-02-02
Feature/2nd kind

mzbush:
Adds class for Chebyshev polynomials of the second kind

### DIFF
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -210,7 +210,10 @@ class ChebyshevSecondKind(BasisFunction):
     def __init__(self,n):
         super().__init__()
         self._n = n
-        self._points = (numpy.polynomial.chebyshev.chebpts2(n+2)[1:-1] if n>1 else [0])
+        if n > 1:
+            self._points = [numpy.cos(k*numpy.pi/(n+1)) for k in range(1, n+1)]
+        else:
+            self._points = [0.]
 
     @property
     def points(self):
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -192,7 +192,7 @@ class ChebyshevSecondKind(BasisFunction):
         U_1(x) = 2x
         U_{n+1}(x) = 2x U_n(x) - U_{n-1}(x)
 
-    The :attr:`points` for this polynomial are the extrema on the domain
+    The :attr:`points` for this polynomial are the zeros on the domain
     :math:`[-1,1]`:
 
     .. math::
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -197,7 +197,7 @@ class ChebyshevSecondKind(BasisFunction):
 
     .. math::
 
-        x_i^* = -\cos(\pi i/n), i = 0,...,n
+        x_i^* = -\cos(\pi i/(n+1)), i = 1,...,n
 
     For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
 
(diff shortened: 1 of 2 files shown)


mphoward (changes requested):
Thanks Michaela! I had one question about the definition of the points for this basis function.

It would be good if you could also move this code up so that it comes after `ChebyshevFirstKind`. You should try to keep similar classes together within a module.

smolyay/basis.py
```diff
@@ -180,6 +180,145 @@ def make_nested_set(cls, exactness):
...
+    The Chebyshev polynomial :math:`U_n` of degree *n* is defined by the
+    recursive relationship:
+
+    .. math::
+
+        U_0(x) = 1
+        U_1(x) = 2x
+        U_{n+1}(x) = 2x U_n(x) - U_{n-1}(x)
+
+    The :attr:`points` for this polynomial are the extrema on the domain
```
mphoward, line 195:
```suggestion
    The :attr:`points` for this polynomial are the zeros on the domain
```

smolyay/basis.py
```diff
@@ -180,6 +180,145 @@ def make_nested_set(cls, exactness):
...
+    :math:`[-1,1]`:
+
+    .. math::
+
+        x_i^* = -\cos(\pi i/n), i = 0,...,n
```
mphoward, line 200:
```suggestion
        x_i^* = -\cos(\pi i/(n+1)), i = 1,...,n
```

smolyay/basis.py
```diff
@@ -180,6 +180,145 @@ def make_nested_set(cls, exactness):
...
+                start_level = 0
+                end_level = 1
+            level_range = range(start_level, end_level+1)
+
+            basis_functions.extend(ChebyshevSecondKind(n) for n in level_range)
+            levels.append(list(level_range))
+            for p in basis_functions[end_level].points:
+                if not numpy.isclose(points, p).any():
+                    points.append(p)
+        points = [0] + points
```
mphoward, line 319: Why do you prepend `[0]` here? Shouldn't it be generated already? This seems hacky.

mphoward (changes requested):

smolyay/basis.py
mzbush: The issue is that the 0th level of the nested group contains 2 functions, the 0th degree and the 1st degree, while for the first kind the 0th level only has the 0th degree. Since it's set up to add points for the last degree in each level, it gets the 0 from 1 but not from 0.
I could replace the [0] on line 319 with ChebyshevSecondKind(0).points or with basis_functions[0].points
I could also do this:

`if i > 0:`
`    start_level = 2**i`
`    end_level = 2**(i+1)-1`
`else:`
`      start_level = 0`
`      end_level = 1`
`      points = ChebyshevSecondKind(0).points`

mphoward: OK, I think I see! Wouldn’t a better option be to handle the i=0 special case outside the loop, then start the loop counter from 1?

mzbush: So initialize with something like this?

`levels = [[0,1]]`
`basis_functions =  [ChebyshevSecondKind(n) for n in levels[0]]`
`points = basis_functions[0].points + basis_functions[1].points`

mphoward: Yes, although you could also just hard code the two basis functions (rather than do list comprehension) to be explicit. You should also add a comment explaining why you are doing this.

mzbush: Done

mphoward (approved):
Looks good to me, thanks Michaela!

cakieslich (approved):

PR #9, opened 2023-02-07
Feature/derivative

mzf0069:
When a function's derivative is provided, the class `GradientSurrogate`"gives a structure for constructing surrogates.
In order to accomplish this, the method `derivative` is added to the class `ChebyshevFirsKind` and the basis matrix is now constructed in a private method in the class `Surrogate`.
Additionally, the class `Surrogate` now has a new method called `derivative` that evaluates the derivative of the surrogate.


smolyay/basis.py
```diff
@@ -72,6 +80,7 @@ class ChebyshevFirstKind(BasisFunction):
     def __init__(self,n):
         super().__init__()
         self._n = n
+        self._derivative = ChebyshevSecondKind(self._n-1)
```
mphoward, line 83: I know I wrote `self._derivative` on Slack, but it might be better to choose a clearer variable name here, since this is not actually the derivative. Maybe:
```suggestion
        self._derivative_polynomial = ChebyshevSecondKind(self._n-1)
```
?

smolyay/basis.py
```diff
@@ -128,6 +137,35 @@ def __call__(self,x):
                 answer += math.comb(self._n,2*k)*((x**2 - 1)**k)*(x**(self._n-2*k))
             return answer
 
+    def derivative(self, x):
```
mphoward, line 140: You should add an abstract `derivative` method to the `BasisFunction` that defines this interface. Then, every basis function is responsible for implementing the method.

smolyay/basis.py
```diff
@@ -128,6 +137,35 @@ def __call__(self,x):
...
+            Value of the derivative of Chebyshev polynomials of first kind.
+
+        Raises
+        ------
+        ValueError
+            if input is outside the domain [-1,1].
+        """
+        if x > 1 or x < -1:
+            raise ValueError("Input is outside the domain [-1,1]")
+        return self.n*self._derivative(x)
```
mphoward, line 167: With the change above:
```suggestion
        return self.n*self._derivative_polynomial(x)
```

smolyay/surrogate.py
```diff
@@ -156,6 +156,118 @@ def grid_generator(self, value):
         self._grid_generator = value
         self._reset_grid()
 
+    def derivative(self, x):
```
mphoward, line 159: This should be called gradient to make it clear that it will return all components for a multivariable function.
```suggestion
    def gradient(self, x):
```

smolyay/surrogate.py
```diff
@@ -156,6 +156,118 @@ def grid_generator(self, value):
...
+        x_scaled = self._mapdomain(x, self._domain, [[-1, 1]]*self.dimension)
+        # evaluate the gradient
+        for ni in range(self.dimension):
+            value = 0
+            for coeff, basis in zip(self.coefficients,
+                                    self.grid.basis_functions):
+                if self.dimension > 1:
+                    term = numpy.prod(
+                        [f.derivative(x) if dim == ni else f(x) for x, f, dim
+                         in zip(x_scaled, basis, range(self.dimension))])
```
mphoward, line 205: This is great! You can achieve the thing with the counter using `enumerate`:
```suggestion
                        [f.derivative(x) if dim == ni else f(x) for dim, (x, f)
                         in enumerate(zip(x_scaled, basis))])
```

smolyay/surrogate.py
```diff
@@ -156,6 +156,118 @@ def grid_generator(self, value):
...
+                else:
+                    term = coeff*basis.derivative(x_scaled[0])
```
mphoward, line 207: I think no factor of `coeff` here because it is on the next line?
```suggestion
                    term = basis.derivative(x_scaled[0])
```

smolyay/surrogate.py
```diff
@@ -156,6 +156,118 @@ def grid_generator(self, value):
...
+                value += coeff*term
+            gradient.append(value)
+
+        return gradient[0] if self.dimension == 1 else gradient
+
+    def _make_basis_matrix(self, gradient=False):
```
mphoward, line 213: Instead of passing a bool `gradient` argument here, you should instead leave `_make_basis_matrix` in its original form here, then override it in `GradientSurrogate` if you want it to do something different for that class.

mphoward (changes requested):
This is a good first implementation! Here are some comments on the code.

smolyay/basis.py
```diff
@@ -310,6 +337,38 @@ def __call__(self,x):
...
+        """
+        if x > 1 or x < -1:
+            raise ValueError("Input is outside the domain [-1,1]")
+        if x == 1:
+            return self._n*(self._n+1)*(self._n+2)/3
+        elif x == -1:
+            return ((-1)**(self._n+1))*self._n*(self._n+1)*(self._n+2)/3
+        else:
+            return ((self._n+1)*ChebyshevFirstKind(self._n+1)(x) -
+                    x*ChebyshevSecondKind(self._n)(x))/(x**2-1)
```
mphoward, line 370: Per our discussion on Slack, you should not make a new polynomial each time. Instead, you can make T_{n+1}, then use the call method of the self for the second kind::
```suggestion
            if self._derivative_polynomial is None:
                self._derivative_polynomial = ChebyshevFirstKind(self._n+1)
            return ((self._n+1)*self._derivative_polynomial(x) -
                    x*self(x))/(x**2-1)
```
You should set `self._derivative_polynomial = None` in the constructor.

You can do a similar `if not None` syntax in the `ChebyshevFirstKind`, if you'd like to delay making the polynomial until the derivative is evaluated. It would make sense to do both in the same way.

mphoward (changes requested):

smolyay/basis.py
```diff
@@ -56,22 +71,31 @@ class ChebyshevFirstKind(BasisFunction):
...
     For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
 
+    The :meth:`derivative` takes the derivative of Chebyshev polynomials
+    of first kind:
+    ..math:
+
+        T_n'(x) = nU_{n-1}(x)
+
+    where :math:`U_n` is Chebyshev polynomials of second kind.
+
```
mphoward, line 88: You can delete this, it is already documented below.
```suggestion
```

smolyay/basis.py
```diff
@@ -202,16 +258,28 @@ class ChebyshevSecondKind(BasisFunction):
...
+    of fsecond kind:
+
+    ..math::
+        U_n'(x) = \frac{(n+1)T_{n+1}(x)-xU_n(x)}{x^{2}-1}
+
+    The above equation does not converge for :math:x={-1, 1}.
+
+    ..math::
+        \lim_{x \to 1} U_n'(x) = \frac{n(n+1)(n+2)}{3}
+        \lim_{x \to -1} U_n'(x) = (-1)^{n+1} \frac{n(n+1)(n+2)}{3}
```
mphoward, line 271: Move this math to the `derivative` method, it will be more helpful there.

smolyay/surrogate.py
```diff
@@ -315,3 +369,128 @@ def _reset_grid(self):
...
+            raise IndexError("Data must be of size of number"
+                             "of grid points x dimension.")
+        points, basis_functions = self.grid.points, self.grid.basis_functions
+        # create basis matrix with appropriate size
+        basis_matrix = numpy.zeros((self.dimension * len(points), len(points)))
+        # evalaute each term
+        count = -1
+        for i, point in enumerate(points):
+            for ni in range(self.dimension):
+                count += 1
```
mphoward, line 482: I would move this below and put after you assign the value to `basis_matrix`. Then, you can start `count = 0`. This would be a more typical pattern.

tests/basis_test.py
```diff
@@ -67,6 +67,58 @@ def test_cheb_call_random_points():
...
+
+
+def test_cheb_derivative_polynomial():
+    """Test if the correct object is created for derivation."""
+    f2 = ChebyshevFirstKind(2)
+    derivative_polynomial = f2._derivative_polynomial
+    f2_derivative = f2.derivative(0.5)
+    assert derivative_polynomial is None
+    assert isinstance(f2._derivative_polynomial, ChebyshevSecondKind)
+    assert numpy.isclose(f2._derivative_polynomial(0.5), 1)
```
mphoward, line 92: You should not test internal details of the class unless absolutely necessary, as these could change. The test of correctness above is sufficient. You could probably just remove this whole test.

tests/basis_test.py
```diff
@@ -67,6 +67,58 @@ def test_cheb_call_random_points():
...
         assert numpy.isclose(f(x),special.eval_chebyt(n,x))
 
+
+def test_cheb_derivative():
+    """Test if the correct derivative is generated."""
+    f0 = ChebyshevFirstKind(0)
+    f1 = ChebyshevFirstKind(1)
+    f2 = ChebyshevFirstKind(2)
+    x0, x1 = 1, -0.5
+    assert numpy.isclose(f0.derivative(x0), 0)
```
mphoward, line 77: You might get nicer output using `pytest.approx` for these comparisons:
```suggestion
    assert f0.derivative(x0) == pytest.approx(0)
```

tests/basis_test.py
```diff
@@ -67,6 +67,58 @@ def test_cheb_call_random_points():
...
+    assert numpy.isclose(u2.derivative(x0), 8)
+    assert numpy.isclose(u0.derivative(x1), 0)
+    assert numpy.isclose(u1.derivative(x1), 2)
+    assert numpy.isclose(u2.derivative(x1), -8)
+    assert numpy.isclose(u0.derivative(x2), 0)
+    assert numpy.isclose(u1.derivative(x2), 2)
+    assert numpy.isclose(u2.derivative(x2), 4)
+
+
+def test_cheb_2nd_derivative_polynomial():
```
mphoward, line 112: Same comments on this as above

tests/basis_test.py
```diff
@@ -67,6 +67,58 @@ def test_cheb_call_random_points():
...
+    x0, x1, x2 = 1, -1, 0.5
+    assert numpy.isclose(u0.derivative(x0), 0)
+    assert numpy.isclose(u1.derivative(x0), 2)
+    assert numpy.isclose(u2.derivative(x0), 8)
+    assert numpy.isclose(u0.derivative(x1), 0)
+    assert numpy.isclose(u1.derivative(x1), 2)
+    assert numpy.isclose(u2.derivative(x1), -8)
+    assert numpy.isclose(u0.derivative(x2), 0)
+    assert numpy.isclose(u1.derivative(x2), 2)
+    assert numpy.isclose(u2.derivative(x2), 4)
```
mphoward, line 109: Long lists like this are hard to read, why not do this instead?
```suggestion
    assert numpy.isclose(u0.derivative(1), 0)
    assert numpy.isclose(u1.derivative(1), 2)
    assert numpy.isclose(u2.derivative(1), 8)

    assert numpy.isclose(u0.derivative(-1), 0)
    assert numpy.isclose(u1.derivative(-1), 2)
    assert numpy.isclose(u2.derivative(-1), -8)

    assert numpy.isclose(u0.derivative(0.5), 0)
    assert numpy.isclose(u1.derivative(0.5), 2)
    assert numpy.isclose(u2.derivative(0.5), 4)
```
Same comment for the other tests

tests/surrogate_test.py
```diff
@@ -46,12 +59,18 @@ def test_initialization_1d():
         ChebyshevFirstKind.make_nested_set(2))
     domain = (-1, 1)
     surrogate = Surrogate(domain, grid_generator)
+    gradient_surrogate = GradientSurrogate(domain, grid_generator)
```
mphoward, line 62: `Surrogate` and `GradientSurrogate` should be tested separately, write individual test functions for each.

tests/surrogate_test.py
```diff
@@ -87,9 +119,13 @@ def test_map_function_1d():
     domain = (-8, 12)
     new_domain = (0, 1)
     f = Surrogate(domain, grid)
-    assert f._mapdomain(2, domain, new_domain) == pytest.approx(0.5)
+    g = GradientSurrogate(domain, grid)
+    assert (f._mapdomain(2, domain, new_domain) and
```
mphoward, line 123: You do not need to retest functions that are simply inherited from the parent class.

tests/surrogate_test.py
```diff
@@ -156,13 +196,43 @@ def test_train_from_data():
     """Test if train_from_data_works."""
     grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(4))
     surrogate = Surrogate([(-5, 10), (0, 15)], grid_gen)
+    gradient_surrogate = GradientSurrogate([(-5, 10), (0, 15)], grid_gen)
```
mphoward, line 199: Again, separate test! And here it is even more important since the `linear_solver` is not a parameter for the `GradientSurrogate`.

mphoward (changes requested):
Thanks for the fixes! Here are some small comments on the code, and some changes requested for the tests.

mphoward (approved):
Looks good to me! If this is also working in your practical tests, I think it's ready to merge. Thanks for the nice work!

PR #10, opened 2023-05-24
Feature/slow growth nested

mzbush:
files for the adaptive slow growth, which has been tested and designed with nested basis functions in mind as per the paper 'Slow Exponential Growth for Clenshaw Curtis Sparse Grids' by John Burkardt


mphoward:
This PR can be revisited with the new API and refactor branch.

PR #12, opened 2023-05-30
Implement faster chebyshev evaluation

mzbush:
Using Python libraries is generally faster

### DIFF
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -2,6 +2,7 @@ import abc
 import math
 
 import numpy
+import scipy.special
 
 class BasisFunction(abc.ABC):
     """Basis function for interpolating data.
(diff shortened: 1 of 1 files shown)


mphoward:
Thanks! Could you please report timings for evaluating the `ChebyshevFirstKind` before and after this PR?

You can use the built-in `timeit` module to do this:
https://docs.python.org/3/library/timeit.html

mzbush:
When running 1000 random calls for n = 0 to 99 each, it took about 4.17 seconds before. With this fix, it takes about 0.33 seconds.

mphoward:
Thanks, that is great! To make it easier for me to understand, can you run just `n=2, 4, 8, 16, 32` for one point (say `x=0.5`), and report the `timeit` for 1000 or maybe even 10000 calls as a table here in the PR with three columns: `n`, old method, new method.

How to make tables of GitHub Markdown:

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables

mzbush:
Times in seconds after running 100,000 times with x = 0.5
 | n| Before  | After
|----|----------|----------|
| 2|0.143598 |0.119076|
| 4|0.201770 |0.120749|
| 8|0.333639 |0.121037|
|16|0.656484 |0.123332|
|32|1.482065 |0.131643|

mphoward:
Great! This is what I thought could happen. The cost of evaluating the polynomials increases greatly with n for the old method, but is nearly constant for the new method. This will be a huge improvement for higher accuracy surrogates that contain higher order polynomials. Good find and good work!

mphoward (approved):
This is a drop-in replacement. If you have verified all existing tests pass, this is ready to merge.

PR #14, opened 2023-06-29
Allow multiple inputs to basis and surrogate

mzbush:
This pull request is meant to allow for the surrogate call to take multiple inputs at once.


smolyay/basis.py
```diff
@@ -131,7 +131,7 @@ def __call__(self, x):
             if input is outside the domain [-1,1]
 
         """
-        if x > 1 or x < -1:
+        if numpy.any([numpy.greater(x,1),numpy.less(x,-1)]):
```
mphoward, line 134: This should avoid an array conversion and is the analog of what was here before:
```suggestion
        if numpy.any(numpy.greater(x, 1)) or numpy.any(numpy.less(x, -1)):
```
If `x` could be required to be a NumPy array, you could just use `x > 1` and `x < -1` instead the functions, but this is a good compatible work around.

Please apply to other spots below.

smolyay/basis.py
```diff
@@ -335,17 +335,23 @@ def derivative(self, x):
...
             raise ValueError("Input is outside the domain [-1,1]")
-        if x == 1:
-            return self._n*(self._n+1)*(self._n+2)/3
-        elif x == -1:
-            return ((-1)**(self._n+1))*self._n*(self._n+1)*(self._n+2)/3
+        if self._derivative_polynomial is None:
+            self._derivative_polynomial = ChebyshevFirstKind(self._n+1)
+        x = numpy.array(x, copy=False)
```
mphoward, line 342:
```suggestion
        x = numpy.asarray(x)
```

smolyay/basis.py
```diff
@@ -335,17 +335,23 @@ def derivative(self, x):
...
+        y = numpy.zeros(x.shape)
+        flag1 = numpy.logical_not(numpy.logical_or(x == 1,x == -1))
+        flag2 = x == 1
+        flag3 = x == -1
```
mphoward, line 346: You should reuse these in the order I suggested so you don't duplicate calculations. Also, I think that it is allowed to use boolean operators, so I would suggest this as easier to read:
```suggestion
        flag2 = x == 1
        flag3 = x == -1
        flag1 = ~(flag2 | flag3)
```

smolyay/basis.py
```diff
@@ -335,17 +335,23 @@ def derivative(self, x):
...
+        y[flag2] = self._n*(self._n + 1)*(self._n + 2)/3
+        y[flag3] = ((-1)**(self._n+1))*self._n*(self._n + 1)*(self._n + 2)/3
+        if y.shape == ():
+            return y[()]
         else:
-            if self._derivative_polynomial is None:
-                self._derivative_polynomial = ChebyshevFirstKind(self._n+1)
-            return ((self._n+1)*self._derivative_polynomial(x) -
-                    x*self(x))/(x**2-1)
+            return numpy.squeeze(y)
```
mphoward, line 354: What is the reason for `squeeze` here?

I also think you can use `y.item()` to return the single scalar value in the if.

smolyay/surrogate.py
```diff
@@ -192,13 +192,23 @@ def gradient(self, x):
...
+            if x.ndim == 2 and x.shape[0] == 1 and x.shape[1]  > 1:
+                # the cast to 2d puts these in wrong order, so transpose
+                x = x.T
+            else:
+                x = x[..., numpy.newaxis]
+        if x.shape[-1] != self.dimension:
+            raise IndexError("Input must match dimension of domain")
+        if self.dimension > 1:
+            oob = any(numpy.any(xi < bound[0]) or numpy.any(xi > bound[1]) 
+                    for xi,bound in zip(x.T,self.domain))
```
mphoward, line 206: Rewrite this to use loop over a counter rather than zip, I think we made this change in the benchmark functions already.

smolyay/surrogate.py
```diff
@@ -192,13 +192,23 @@ def gradient(self, x):
...
+        else:
+            oob = (numpy.any(x < self.domain[0]) or 
+                    numpy.any(x > self.domain[1]))
+        if oob:
             raise ValueError('x must lie in domain of surrogate')
-
         gradient = []
```
mphoward, line 212: Shouldn't `gradient` be a NumPy array rather than a list? You could preallocate it here, then assign the values below for each component.

smolyay/surrogate.py
```diff
@@ -209,14 +219,16 @@ def gradient(self, x):
...
-                    term = basis.derivative(x_scaled[0])
+                    term = basis.derivative(x_scaled)
+                
                 value += coeff*term
-            gradient.append(value)
-
-        return gradient[0] if self.dimension == 1 else gradient
+            gradient.append(numpy.squeeze(value))
+        gradient = numpy.array(gradient,copy=False).transpose()
+        return numpy.squeeze(gradient)
```
mphoward, line 231: I don't follow what all is going on here, could you give an example of why you need this order of squeeze and transpose? Maybe some of this could be fixed using the suggestion to preallocate above.

smolyay/surrogate.py
```diff
@@ -243,24 +255,35 @@ def __call__(self, x):
...
+            if x.ndim == 2 and x.shape[0] == 1 and x.shape[1]  > 1:
+                # the cast to 2d puts these in wrong order, so transpose
+                x = x.T
+            else:
+                x = x[..., numpy.newaxis]
+        if x.shape[-1] != self.dimension:
+            raise IndexError("Input must match dimension of domain")
+        if self.dimension > 1:
+            oob = any(numpy.any(xi < bound[0]) or numpy.any(xi > bound[1]) 
+                    for xi,bound in zip(x.T,self.domain))
```
mphoward, line 269: Same comment here

smolyay/surrogate.py
```diff
@@ -470,32 +493,32 @@ def train_from_data(self, data):
...
+            for j, basis in enumerate(basis_functions):
+                if self.dimension > 1:
+                    value = numpy.prod(
+                            [basis[dim].derivative(points[...,dim]) 
+                             if dim == ni else basis[dim](points[...,dim]) 
+                             for dim in range(len(basis))],axis=0)
+                else:
+                    value = basis.derivative(points)
+                basis_matrix[ni + self.dimension*numpy.array(range(num_points)), 
+                              j] = value
```
mphoward, line 521: There must be a better way to write this assignment using slicing. Can you explain what you are doing?

mphoward (changes requested):
This is a good first pass! I commented on the main code, and we can have a look at the tests later.

smolyay/surrogate.py
mzbush: The way this was originally written, the row index was determined by the variable `count` which went up by 1 through the for loops for the points and the dimensions.

```
        for i, point in enumerate(points):
            for ni in range(self.dimension):
                count += 1
```

My goal was to remove the loop for the points, so I needed a way to figure out the value of count so that all the elements in `value` get assigned to the right place in `basis_matrix`.

The value of `count` can be calculated at every step by `ni + self.dimension*i`. Removing that first for loop means that you're calculating all values that go in indexes `ni + self.dimension*i for i in range(0, len(points))`. What's currently written is a more compact way of getting all those indexes. These indexes aren't next to each other (for example when ni = 1 and self.dimension = 3 the indexes are 1, 4, 7, 10, 13, ... etc) which makes it more complicated.

smolyay/basis.py
mzbush: Squeeze is just my usual go-to for dimension reducing. It just wasn't adequate in this particular case which also made it unnecessary.

smolyay/surrogate.py
mphoward: I see, so if you want to assign to indexes starting from `ni` in steps of `self.dimension`, you can use standard Python slicing:
```
basis_matrix[ni::self.dimension, j] = value
```

smolyay/surrogate.py
mzbush: I was replicating the way is was originally written and so the value is squeezed to avoid a nested array being added to gradient. 

The transpose is added because gradient is supposed to be a matrix with a size num_points x dimension. value will be an array the size of num_points. However, appending adds value to gradient's row when it should be the column, hence the transpose.

The last squeeze is to again collapse the number of dimensions, which is particularly important in the 1D case. Originally the return line was 

`return gradient[0] if self.dimension == 1 else gradient`

which had the effect of preventing gradient from being a nested list for 1D cases. However, this obviously would prevent any gradients being returned for any points except the first one, so it was replaced with squeeze which did the same thing.

The transpose was causing issues if the points were given in multi-dimensional grids (so gradient's shape had more than 2 dimensions), and I pushed a commit fixes this issue.

mzbush: I allocated gradient to a numpy array with removes the need for reshaping completely, but both instances of squeezing are required to pass all the pytests.

smolyay/surrogate.py
mzbush: It is done.

smolyay/surrogate.py
```diff
@@ -192,31 +192,43 @@ def gradient(self, x):
...
+                             for dim in range(len(basis))], axis=0)
                 else:
-                    term = basis.derivative(x_scaled[0])
+                    term = basis.derivative(x_scaled)
+                
                 value += coeff*term
-            gradient.append(value)
-
-        return gradient[0] if self.dimension == 1 else gradient
+            gradient[...,ni] = numpy.squeeze(value)
```
mphoward, line 230: I think it should be sufficient to reshape this
```suggestion
            gradient[...,ni] = numpy.reshape(value, gradient.shape[:-1])
```

mphoward (changes requested):
Thanks! Here are a couple more small things. I haven't looked at the tests themselves yet.

I am still not totally sure we should always be calling `squeeze`. I think there are cases where this converts the output to the wrong shape. We should probably discuss what types of inputs are possible, then list what the right type of output should be & decide if squeeze correctly implements all of these cases. We could write a helper function that helps cast output for use in all of these functions if necessary.

smolyay/basis.py
```diff
@@ -161,7 +161,7 @@ def derivative(self, x):
         ValueError
             if input is outside the domain [-1,1].
         """
-        if x > 1 or x < -1:
+        if numpy.any([numpy.greater(x,1),numpy.less(x,-1)]):
```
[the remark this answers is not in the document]
mzbush: Done.

mphoward:
This PR is in very good shape. The only thing that remains to be done is to list out all the possible scenarios of inputs to the functions. Here are the ones I can think of quickly, with the mocked up `input -> output`:

* Dimensionality 1:
  - `1` -> `100`
  - `[1]` -> `np.array([100])`
  - `[[1]]` -> `np.array([100])`
  - `[1, 2]` -> `np.array([100, 200])`
  - `[[1], [2]]` -> `np.array([100, 200])`
  - `np.ones((1, 1, 1, 1))` -> `np.full((1, 1, 1), 100)`
  - `np.ones((1, 2, 3, 4))` -> `np.full((1, 2, 3, 4), 100)`
  - `np.ones((4, 3, 2, 1))` -> `np.full((4, 3, 2), 100)`
* Dimensionality > 1 (say, 3d):
  - `[1, 1, 1]` -> `100`
  - `[[1, 1, 1]]` -> `np.array([100])`
  - `[[1, 1, 1], [2, 2, 2]]` -> `np.array([100, 200])`
  - `np.ones((1, 1, 1, 3))` -> `np.full((1, 1, 1), 100)`
  - `np.ones((4, 3, 2, 3))` -> `np.full((4, 3, 2), 100)`

You should make sure that the unit tests work in the appropriate scenarios for the basis functions and the surrogate. I suspect that `squeeze` is going to cause some to fail.

smolyay/surrogate.py
```diff
@@ -284,7 +303,11 @@ def __call__(self, x):
             else:
                 term = basis(x_scaled)
             value += coeff*term
-        return numpy.squeeze(value)
+        if output_scalar:
+            value = value.item()
+        else:
+            value = numpy.reshape(value,input_shape)
```
mphoward, line 309: Can you apply similar logic to what I had above here? For example, I think you can only get a scalar if your input shape was 0d (for a 1d problem) or 1d (for a >= 2d problem). Otherwise, you should discard the last dimension of the input?
```suggestion
        if (self.dimension == 1 and input_shape.ndim == 0) or (self.dimension > 1 and input_shape.ndim == 1):
            value = value.item()
        else:
            value = numpy.reshape(value, input_shape[:-1])
```

smolyay/surrogate.py
```diff
@@ -192,13 +192,20 @@ def gradient(self, x):
...
         if self._coefficients is None:
             raise ValueError('Function needs training!')
+        input_shape = numpy.shape(x)
         x = numpy.array(x, copy=False, ndmin=2)
         if self.dimension == 1:
             if x.ndim == 2 and x.shape[0] == 1 and x.shape[1]  > 1:
                 # the cast to 2d puts these in wrong order, so transpose
                 x = x.T
-            else:
+            elif x.shape[-1] > 1:
```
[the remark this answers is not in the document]
mzbush: Yes. In the case of an input shaped like (4,3,2,1), adding an extra dimension is unnecessary. Adding a dimension is explicitly for the case where the x.shape[-1] != self.dimension check would fail when self.dimension is 1, which should never happen because a 1D function should always pass that check.

smolyay/surrogate.py
```diff
@@ -227,8 +234,13 @@ def gradient(self, x):
...
                 
                 value += coeff*term
-            gradient[...,ni] = numpy.squeeze(value)
-        return numpy.squeeze(gradient)
+            gradient[...,ni] = numpy.reshape(value,gradient.shape[:-1])
+
+        if output_scalar:
+            gradient = gradient.item()
+        else:
+            gradient = numpy.reshape(gradient,input_shape)
```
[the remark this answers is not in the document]
mzbush: input_shape doesn't have an attribute ndim because it's a tuple, but I get what you mean.

smolyay/surrogate.py
```diff
@@ -192,13 +192,20 @@ def gradient(self, x):
         """
         if self._coefficients is None:
             raise ValueError('Function needs training!')
+        input_shape = numpy.shape(x)
         x = numpy.array(x, copy=False, ndmin=2)
         if self.dimension == 1:
             if x.ndim == 2 and x.shape[0] == 1 and x.shape[1]  > 1:
```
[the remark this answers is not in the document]
mzbush: I feel that it doesn't really work and the way it is currently is more readable. The issue the transpose is trying to solve is, after all, having to do with its current shape, not what it was before.

smolyay/surrogate.py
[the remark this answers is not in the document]
mphoward: Oops, yes use `len(input_shape)`

smolyay/surrogate.py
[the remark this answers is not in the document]
mzbush: I honestly didn't think it would work but when I checked the tests it appears to work fine.

smolyay/surrogate.py
```diff
@@ -192,31 +192,52 @@ def gradient(self, x):
...
+        input_shape = numpy.shape(x)
+        x = numpy.array(x, copy=False, ndmin=2)
+        if self.dimension == 1:
+            if len(input_shape) <= 1:
+                # the cast to 2d puts these in wrong order, so transpose
+                x = x.T
+            elif x.shape[-1] > 1:
+                x = x[..., numpy.newaxis]
+            elif input_shape == x.shape:
+                input_shape = input_shape[:-1]
```
mphoward, line 204: This is not related to getting `x` in the right shape, so it should be removed here. Its purpose is only for output, handle below.
```suggestion
```

smolyay/surrogate.py
```diff
@@ -192,31 +192,52 @@ def gradient(self, x):
...
                 else:
-                    term = basis.derivative(x_scaled[0])
+                    term = basis.derivative(x_scaled)
+                
                 value += coeff*term
-            gradient.append(value)
+            gradient[...,ni] = numpy.reshape(value, gradient.shape[:-1])
 
-        return gradient[0] if self.dimension == 1 else gradient
+        if input_shape == ():
```
mphoward, line 236: Per my first review:
```suggestion
        if len(input_shape) == 0:
```
The gradient is only scalar for 1d problems with a scalar input, which is the only case where the input shape has length 0. In any other case, the shape of the gradient must match the input shape so the second condition is fine.

smolyay/surrogate.py
```diff
@@ -243,23 +264,44 @@ def __call__(self, x):
...
-                term = basis(x_scaled[0])
+                term = basis(x_scaled)
             value += coeff*term
 
+        if (self.dimension == 1 and input_shape == ()) or (self.dimension > 1 and len(input_shape) == 1):
+            value = value.item()
+        elif (self.dimension == 1 and (len(input_shape) == 1 or input_shape[-1] != 1)):
+            value = numpy.reshape(value, input_shape)
+        else:
+            value = numpy.reshape(value, input_shape[:-1])
```
mphoward, line 304: I think this would a little simpler and easier to read?
```suggestion
        if (self.dimension == 1 and len(input_shape) == 0) or (self.dimension > 1 and len(input_shape) == 1):
            value = value.item()
        else:
            if self.dimension == 1 and (len(input_shape) <= 1 or input_shape[-1] > 1):
                # an extra dimension was added in these cases, so use original input shape
                output_shape = input_shape
            else:
                # evaluation of the function drops the last dimension from x
                output_shape = x.shape[:-1]
            value = numpy.reshape(value, output_shape)
```

mphoward (changes requested):

smolyay/surrogate.py
```diff
@@ -304,14 +350,14 @@ def train_from_data(self, data, linear_solver='lu'):
 
         # make basis matrix
         points, basis_functions = self.grid.points, self.grid.basis_functions
+        points = numpy.array(points)
```
mphoward, line 353:
```suggestion
        points, basis_functions = numpy.array(self.grid.points), self.grid.basis_functions
```

smolyay/surrogate.py
```diff
@@ -470,32 +516,31 @@ def train_from_data(self, data):
...
         self._data = numpy.array(data, ndmin=1)
         expected_data_shape = ((len(self.grid.points), self.dimension)
                                if self.dimension > 1 else (
                                        len(self.grid.points),))
         if self._data.shape != expected_data_shape:
-            raise IndexError("Data must be of size of number"
+            raise IndexError("Data must have size of number"
                              "of grid points x dimension.")
         points, basis_functions = self.grid.points, self.grid.basis_functions
+        points = numpy.array(points)
```
mphoward, line 529:
```suggestion
        points, basis_functions = numpy.array(self.grid.points), self.grid.basis_functions
```

tests/basis_test.py
```diff
@@ -67,6 +67,14 @@ def test_cheb_call_random_points():
         f = ChebyshevFirstKind(n)
         assert numpy.isclose(f(x),special.eval_chebyt(n,x))
 
+def test_cheb_call_random_points_multi_input():
```
mphoward, line 70: You don't need to run this test on multiple `ns`. Just choose one, unless there are special cases and then include those explicitly.

tests/basis_test.py
```diff
@@ -170,22 +190,36 @@ def test_cheb_2nd_call_random_points():
         f = ChebyshevSecondKind(n)
         assert numpy.isclose(f(x),special.eval_chebyu(n,x))
 
+def test_cheb_2nd_call_random_points_multi_input():
```
mphoward, line 193: Same here.

tests/surrogate_test.py
```diff
@@ -180,6 +180,53 @@ def test_surrogate_1():
...
+    assert numpy.allclose(surrogate_values, surrogate_values_multi)
+
+
+def test_surrogate_1_multi_grid():
+    """Test if surrogate generates exact results if call has >1 input."""
+    grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(2))
+    surrogate = Surrogate([(-1, 1), (-1, 1)], grid_gen)
+    surrogate.train(function_1)
+    # random points in the domain
+    points = numpy.random.rand(3,4,2)*2 - 1
```
mphoward, line 201: Use of `numpy.random` in tests is discouraged because it makes them not reproducible.

If you want to do this, make sure you seed it each time. Using the modern NumPy workflow for creating a RNG will help with this, rather than relying on `numpy.random`. See the docs there or ask if you have questions.

tests/surrogate_test.py
```diff
@@ -180,6 +180,53 @@ def test_surrogate_1():
...
+    points_list = [[1, 1], [0.5, 0.5]]
+    point_grid = numpy.ones((1,1,1,2))
+    points_grid_1 = numpy.ones((4,3,2,2))
+    points_grid_2 = numpy.ones((4,3,5,2))
+    assert numpy.isscalar(surrogate(point_scalar))
+    assert surrogate(point_nested_list).shape == (1,)
+    assert surrogate(points_list).shape == numpy.shape(points_list)[:-1]
+    assert surrogate(point_grid).shape == (1,1,1)
+    assert surrogate(points_grid_1).shape == (4,3,2)
+    assert surrogate(points_grid_2).shape == (4,3,5)
```
mphoward, line 227: These tests will be easier to read if you just put the variables inside the `surrogate(...)`. Also on line 224, hard code the expected shape, like the others.

tests/surrogate_test.py
```diff
@@ -204,6 +251,52 @@ def test_surrogate_2(linear_solver):
...
+
+@pytest.mark.parametrize('linear_solver', ['lu', 'lstsq', 'inv'])
+def test_surrogate_2_multi_grid(linear_solver):
+    """Test if surrogate generates exact results for 1D function and >1 input"""
+    grid_gen = SmolyakGridGenerator(ChebyshevFirstKind.make_nested_set(2))
+    surrogate = Surrogate((-10, 10), grid_gen)
+    surrogate.train(function_2, linear_solver)
+    # random point in the domain
+    point = numpy.random.rand(3,4)
+    assert numpy.allclose(surrogate(point), function_2(point))
```
mphoward, line 273: Maybe combine these 2 tests, since there are only 2 new additional lines in it?

tests/surrogate_test.py
```diff
@@ -204,6 +251,52 @@ def test_surrogate_2(linear_solver):
...
+    points_grid_1 = numpy.ones((1,2,3,4))
+    points_grid_2 = numpy.ones((4,3,2,1))
+    assert numpy.isscalar(surrogate(point_scalar))
+    assert surrogate(point_list_scalar).shape == numpy.shape(point_list_scalar)
+    assert surrogate(point_nested_list_scalar).shape == (1,)
+    assert surrogate(points_list).shape == numpy.shape(points_list)
+    assert surrogate(points_nested_list).shape == (2,)
+    assert surrogate(point_grid).shape == (1,1,1)
+    assert surrogate(points_grid_1).shape == (1,2,3,4)
+    assert surrogate(points_grid_2).shape == (4,3,2)
```
mphoward, line 297: Same comment here about making this more readable by putting the arguments inside `surrogate(...)` and hard coding the expected shapes.

tests/surrogate_test.py
```diff
@@ -243,6 +336,53 @@ def test_gradient_surrogate_3():
     assert numpy.allclose(surrogate_gradient_values, exact_values)
 
 
+def test_gradient_surrogate_3_multi_input():
```
mphoward, line 339: Translate comments above down here

mphoward (changes requested):
Looking good! Some changes are needed on the tests, but then this is ready to go.

PR #15, opened 2024-01-16
Fix error in s266 from different size of inputs

mzbush:
Discovered a glitch in which the function s266 where an input matrix with more than 2 axes would error. This bug has been fixed and a new pytest has been added to check all benchmark functions for this glitch as a preventive measure.


tests/benchmark_test.py
```diff
@@ -214,3 +214,11 @@ def test_functions_call_error_above(fun):
     with pytest.raises(ValueError):
         fun(above_bounds)
 
+@pytest.mark.parametrize("fun",functions[:3])
```
mphoward, line 217: Why is this list being sliced to only the first 3 functions?

tests/benchmark_test.py
```diff
@@ -214,3 +214,11 @@ def test_functions_call_error_above(fun):
...
+def test_functions_call_shape(fun):
+    """Test all functions give the correct shape"""
+    x_list = numpy.zeros((2,50,fun.dimension))
+    mid_point = numpy.divide(fun.upper_bounds,2)
+    x_list[0,:,:] = numpy.linspace(fun.lower_bounds,mid_point,50)
```
mphoward, line 222: I'm unsure about how this works because you are assigning a 1d array into a 2d slice, so I would think that is an error. Here's what happens when I run a similar snippet in my interpreter:
```
>>> import numpy as np
>>> x_list = np.zeros((2, 50, 4))
>>> x_list[0,:,:] = np.linspace(0, 1, 50)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: could not broadcast input array from shape (50,) into shape (50,4)
```
Maybe I am doing something wrong here?

mphoward (commented):
Thanks Michaela! The fix looks clean, but I had some questions about the test you added.

tests/benchmark_test.py
mzbush: That is because it shouldn't. That was done to make sure the test worked and then wasn't removed.

tests/benchmark_test.py
mzbush: The bounds of the function are a list with as many elements as the benchmark function has dimensions. When you input the bounds, you get an array of size `(50, fun.dimension)`, where each column is a linspace for each bound. The glitch arose when trying to pass an array of size `(5, 5000, fun.dimension)` into s266, even though all benchmark function are supposed to be able to handle an input of any size as long as the last axis is the same as `fun.dimension`.

`np.linspace(fun.lower_bounds,fun.upper_bounds)` gets you an array of size `(50, fun.dimension)`, exactly the size needed for `x_list[0,:,:]`

mzbush: Done.

mphoward (approved):
Looks good to me, thanks!

PR #16, opened 2024-02-29
Feature/adaptive

mzbush:
Code for creating an adaptive smolyak grid that allows for arbitrary k indexes.


mphoward:
Thanks for this PR! I tried to review the code, but I am slightly confused on this use case. Can you explain more how this class should be used?

mzbush:
> Thanks for this PR! I tried to review the code, but I am slightly confused on this use case. Can you explain more how this class should be used?

The new class is a way to generate a grid with a custom set of smolyak k sets instead of being limited to the exactness specified by the nested basis set. It requires the custom set during initialization and asks for the dimension when called to make it compatible with how the surrogate objects call and use the grid generator classes.

mphoward:
Does that mean that someone would have to specify the entire list of k sets though? That seems pretty clunky. I would think that adaptive sampling would add/remove k sets from some base case.

mphoward:
This PR can be revisited with the new API and algorithms.

PR #19, opened 2024-04-10
Refactor basis functions

mzbush:
Pull request for refactoring the way that the basis functions and points are handled, namely decoupling them from each other so they are not so tightly interlinked.


smolyay/basis.py
```diff
@@ -1,29 +1,33 @@
...
-    @abc.abstractmethod
-    def points(self):
-        """list: Sampling points for interpolation."""
-        pass
+    def natural_domain(self):
+        """list: Domain the sample points come from."""
+        return self._natural_domain
 
     @abc.abstractmethod
     def __call__(self, x):
```
mzbush, line 24: In the benchmark base class, it had the `__call__` method that would take care of sizing issues as well as check if the inputs were in the domain, and then it would return the answer of an abstract `_function` class. Do you think we should do something similar here?

And on that note, the input of the surrogate `__call__ ` and `train_from_data` also have sizing issues to take care of in addition to normalization. Should taking care of the sizing, normalization, and detecting invalid inputs be the job of the base class?

smolyay/samples.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        print('here')
+        points = [0]
+        degree = 0
+        counter = 0
+        while len(points) < num_points:
+            counter = counter + 1
+            degree = 2**counter
+            new_points = -numpy.cos(numpy.pi * numpy.linspace(0, degree, degree + 1) / degree)
+            for p in new_points:
```
mzbush: In theory, it should be possible to use numpy matrix math and the equation used by `numpy.isclose` to get a boolean list that shows exactly which points in `new_points` are new. However, I don't know if doing that would be faster.

smolyay/basis.py
```diff
@@ -1,29 +1,33 @@
...
+    can be evaluated via :meth:`derivative`.
+
+    Parameters
+    ----------
+    natural_domain : list of two numbers
+        the natural domain of the function within which calls are made.
     """
 
-    def __init__(self):
+    def __init__(self, natural_domain):
```
mphoward: Two things:
1. I think this would be better called `domain`.
2. I would make this an abstract property that has to be overridden by deriving classes like:
```py
@property
@abc.abstractmethod
def domain(self):
    pass
```

smolyay/basis.py
mphoward: Hmm, good thought! Yes, I think I would do a concrete `__call__` (and `derivative`) method in the base class, and have `_function` (and `_derivative`) abstract methods that deriving classes must override.

In the `__call__` method,  we can make a decision whether to throw an error if the inputs are not in the right domain or if it is a contract with the caller that the inputs are valid. There is a performance hit to error checking, but the error messages downstream will be easier for the caller to understand. I am good either way.

I don't think we should do any casting or reshaping of `x` in `__call__` before forwarding to `_function` because the `_function` should make a decision about whether that is needed (e.g., the Chebyshev function will just call scipy).

We do expect that `__call__` will return an output consistent with the shape of `x` (a scalar if `x` is a number, otherwise a NumPy array with the matching shape). This logic should be included here.

smolyay/basis.py
```diff
@@ -57,6 +61,23 @@ def derivative(self, x):
         """
         pass
 
+    def outside_domain(self, x):
```
mphoward: Methods should be actions, and I think this is clearer:
```suggestion
    def check_in_domain(self, x):
```
Also, is this part of the public interface to this class? If not, prepend a `_` to indicate it is private.

smolyay/basis.py
```diff
@@ -57,6 +61,23 @@ def derivative(self, x):
...
+        x : float, numpy:ndarray
+            One-dimensional points.
+
+        Returns
+        -------
+        bool
+            True if input was outside domain, False otherwise
+        """
+        return numpy.any(numpy.greater(x, self.natural_domain[1])) or numpy.any(
+            numpy.less(x, self.natural_domain[0])
```
mphoward: I suggest this is done element wise, and allow the caller to decide whether they take `all` or not. These should also be >= and <= since the domains are defined to be closed intervals.
```suggestion
        return numpy.logical_and(numpy.greater_equal(x, self.domain[0]),
            numpy.less_equal(x, self.domain[1]))
```

If you think it would be convenient, you could add a method like:
```py
def check_all_in_domain(self, x):
    return numpy.all(self.check_in_domain(x))
```

smolyay/basis.py
```diff
@@ -70,14 +91,9 @@ class ChebyshevFirstKind(BasisFunction):
...
-    :math:`[-1,1]`:
-
-    .. math::
-
-        x_i^* = -\cos(\pi i/n), i = 0,...,n
-
-    For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
+    The Chebyshev polynomials form an orthonormal basis on the domain 
+    :math:`[-1,1]`, and thus that is their natural domain when used as a basis
+    function.
```
mphoward:
```suggestion
    Their domain is defined to be :math:`-1 \le x \le 1`.
```

smolyay/basis.py
```diff
@@ -128,13 +134,12 @@ def __call__(self, x):
...
         ValueError
-            if input is outside the domain [-1,1]
+            if input is outside the domain [-1, 1]
 
         """
-        if numpy.any(numpy.greater(x, 1)) or numpy.any(numpy.less(x, -1)):
-            raise ValueError("Input is outside the domain [-1,1]")
-
-        return scipy.special.eval_chebyt(self._n,x)
+        if self.outside_domain(x):
```
mphoward: This would be how you might use this:
```suggestion
        if not numpy.all(self.check_in_domain(x)):
```
or else
```py
        if not self.check_all_in_domain(x):
```

smolyay/basis.py
```diff
@@ -159,67 +164,11 @@ def derivative(self, x):
...
-
-            basis_functions.extend(ChebyshevFirstKind(n) for n in level_range)
-            levels.append(list(level_range))
-            for p in basis_functions[end_level].points:
-                if not numpy.isclose(points, p).any():
-                    points.append(p)
-        return NestedBasisFunctionSet(points, basis_functions, levels)
+        if self.outside_domain(x):
+            raise ValueError("Input is outside the domain [-1, 1]")
+        return self.n * scipy.special.eval_chebyu(self._n - 1, x)
```
mphoward:
```suggestion
        return self.n * scipy.special.eval_chebyu(self.n - 1, x)
```

smolyay/basis.py
```diff
@@ -234,14 +183,9 @@ class ChebyshevSecondKind(BasisFunction):
...
 
-    The :attr:`points` for this polynomial are the zeros on the domain
-    :math:`[-1,1]`:
-
-    .. math::
-
-        x_i^* = -\cos(\pi i/(n+1)), i = 1,...,n
-
-    For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
+    The Chebyshev polynomials form an orthonormal basis on the domain
```
mphoward: Edit as for first kind

smolyay/basis.py
```diff
@@ -86,18 +102,8 @@ class ChebyshevFirstKind(BasisFunction):
     """
 
     def __init__(self, n):
-        super().__init__()
+        super().__init__([-1, 1])
```
mphoward: This would move to the `domain` property

smolyay/basis.py
```diff
@@ -86,18 +102,8 @@ class ChebyshevFirstKind(BasisFunction):
...
         self._n = n
```
mphoward: Does `n` need to be readonly now? It seems like the class will work fine if you modify n. If so, you might want to just do `self.n = n` and remove the property below.

smolyay/basis.py
```diff
@@ -328,141 +262,137 @@ def derivative(self, x):
...
-        y[flag1] =  ((self._n+1)*self._derivative_polynomial(x[flag1]) -
-                     x[flag1]*self(x[flag1]))/(x[flag1]**2-1)
-        y[flag2] = self._n*(self._n + 1)*(self._n + 2)/3
-        y[flag3] = ((-1)**(self._n+1))*self._n*(self._n + 1)*(self._n + 2)/3
+        y[flag1] = (
+            (self._n + 1) * scipy.special.eval_chebyt(self._n + 1, x[flag1])
+            - x[flag1] * self(x[flag1])
+        ) / (x[flag1] ** 2 - 1)
+        y[flag2] = self._n * (self._n + 1) * (self._n + 2) / 3
+        y[flag3] = ((-1) ** (self._n + 1)) * self._n * (self._n + 1) * (self._n + 2) / 3
```
mphoward: We should recycle the evaluation on line 283 in line 284, since 284 is 283 times a minus sign. Instead of raising -1 to the power of n, you should instead check whether n is odd (-1) or even (+1) to determine the sign.

smolyay/samples.py
```diff
@@ -0,0 +1,204 @@
...
+        the natural domain of the function within which calls are made.
+    """
+
+    def __init__(self, natural_domain):
+        self._natural_domain = natural_domain
+        self._points = []
+        self._number_points = 0
+
+    @property
+    def natural_domain(self):
```
mphoward: I think you want `domain` again here
```suggestion
    def domain(self):
```

smolyay/samples.py
```diff
@@ -0,0 +1,204 @@
...
+            new_points = -numpy.cos(
+                numpy.pi * numpy.linspace(0, degree, degree + 1) / degree
+            )
+            for p in new_points:
+                if not numpy.isclose(points, p, rtol=0, atol=1e-11).any():
+                    points.append(p)
+        return points
+
+
+class ChebyshevRoots(UnidimensionalPointSet):
```
mphoward: We don't currently use these, so let's not add these points to our code until we have a reason for them.

smolyay/samples.py
mphoward: Given the numerical difficulties we've had with this, I think we should use integer math to figure out which new points are being added at each level so this is exact. We can then convert the integers to points once they are unique.

However, I have a question: can't the Clenshaw-Curtis points be defined for arbitrary N, and not require this nested formalism? I think we might want to come up with a way to distinguish this, and that perhaps includes setting levels. Like, maybe the default constructor here just takes an `N` and makes the normal CC points.

Then, we can make a nested version that derives from this according to the various scheme, and that takes an exactness / level in its constructor. Kinda similar to how we were using the classmethod in the old version.

mphoward (changes requested):
Thanks for the great first pass on this! Here are some comments. I stopped before the tests so that we can nail down the interface first.

smolyay/basis.py
mzbush: By replacing -1 with numpy.sign(x[flag2]) and flag2 = numpy.abs(x) == 1, I can get the entire evaluation in one equation. Using modulo operations to check for even or odd results in either a 0 or a 1, which requires more work to make a -1 and 1, and trying to use modulo ended up being more cumbersome than raising 1 or -1 to some power.

smolyay/basis.py
mzbush: After you put the BasisFunction in a set, there's no reason to change n after initialization. So making it readonly protects it from being changed. If that's not very pythonic, then we can change it.

mphoward: Someone should have enough common sense not to change `n` in that case, so I vote to make it mutable.

smolyay/samples.py
mzbush: I used the greatest common factor to prune out the duplicate points before they were calculated.

Clenshaw-Curtis points can be defined by any of the basis functions that have the nested extrema. The order is just the order that the default exponential growth uses.

I have been thinking about potentially having different orders. The Clenshaw-Curtis Linear growth that adds 2 points on every level besides the first can't use the exponential growth order because then the points added every level wouldn't be symmetrical. If you remove all the positive points like this:

[0, -1.0, -0.7071067811865476, -0.9238795325112867, -0.38268343236508984]

you can see an "essential" order that the points are added in. Then you can decide where you want the level boundaries to be and add the positive values to make all the levels symmetrical.

So while I've never seen the Clenshaw-Curtis Linear growth, I imagine it divides things up like this:

[0] [-1.0, 1.0] [-0.7071067811865476, 0.7071067811865476] [-0.9238795325112867, 0.9238795325112867][-0.38268343236508984, 0.38268343236508984]

mphoward: I think I see what you’re saying, but I was thinking that you can define these points even for a highest degree polynomial that is non-nested?

https://en.m.wikipedia.org/wiki/Clenshaw%E2%80%93Curtis_quadrature

Having these points would be important, for example, to get access to more tensor grids other than those that match the nested criteria.

mzbush: Would you want that to be the generate_points method and then the Clenshaw-Curtis has children that do use levels override it, or do you want generating the points in a non-nested order to be a separate method?

mzbush: I've added a child to the ClenshawCurtisPointSet called the NestedClenshawCurtisPointSet, which uses the generate_points that its parent used to have. The ClenshawCurtisPointSet now generates N extrema for a polynomial of degree N - 1. 

I've also gotten rid of the number_points property so now the user is expected to make points using the generate_points method.

smolyay/basis.py
```diff
@@ -1,34 +1,81 @@
...
 
-    def __init__(self):
-        self._points = []
-
     @property
     @abc.abstractmethod
-    def points(self):
-        """list: Sampling points for interpolation."""
+    def domain(self):
+        """list: Domain the sample points come from."""
```
mphoward: In refactoring our code, I would vote that we prefer to use NumPy arrays as much as possible, with exceptions for cases where it is not possible (e.g., lists of lists of unequal size)
```suggestion
        """numpy.ndarray: Domain the sample points come from."""
```

smolyay/basis.py
```diff
@@ -1,34 +1,81 @@
...
+        Returns
+        -------
+        float
+            Value of the derivative of the basis function.
+        """
+        if not numpy.all(self.check_in_domain(x)):
+            raise ValueError("Input is outside the domain")
+        return self._derivative(x)
+
+    def check_in_domain(self, x):
```
mphoward: I know I suggested the original name for this, but seeing how you're using it, I think `in_domain` might be better:
```suggestion
    def in_domain(self, x):
```

smolyay/basis.py
```diff
@@ -87,24 +127,14 @@ class ChebyshevFirstKind(BasisFunction):
...
-            self._points = [0]
+        self.n = n
 
     @property
-    def points(self):
-        """list: Sampling points at extrema of polynomial."""
-        return self._points
+    def domain(self):
+        """list: Domain the sample points come from."""
+        return [-1, 1]
```
mphoward: If we switch to NumPy array as the return type, this would need to become like this:
```suggestion
        return np.array([-1, 1])
```

smolyay/basis.py
```diff
@@ -328,141 +278,138 @@ def derivative(self, x):
...
+            (self.n + 1) * scipy.special.eval_chebyt(self.n + 1, x[flag1])
+            - x[flag1] * self(x[flag1])
+        ) / (x[flag1] ** 2 - 1)
+        y[flag2] = (
+            (numpy.sign(x[flag2]) ** (self.n + 1))
+            * self.n
+            * (self.n + 1)
+            * (self.n + 2)
+            / 3
+        )
```
mphoward: I think you've sacrificed a lot of readability here. I would do like this:
```suggestion
        u_limit = self.n * (self.n + 1) * (self.n + 2) / 3
        flag_upper = x == 1
        y[flag_upper] = u_limit
        
        flag_lower = x == -1
        y[flag_lower] = (-1) ** (self.n+1) * u_limit

        flag = ~(flag_upper | flag_lower)
        y[flag] = (
            (self.n + 1) * scipy.special.eval_chebyt(self.n + 1, x[flag])
            - x[flag] * scipy.special.eval_chebyu(self.n, x[flag])
        ) / (x[flag] ** 2 - 1)
```

smolyay/basis.py
```diff
@@ -328,141 +278,138 @@ def derivative(self, x):
...
-        """list: List of lists of indexes for points/functions at each level.
         Raises
         ------
-        IndexError
-            max index must be less than total number of points.
+        ValueError
+            If input is outside the domain `[0, 2\pi]`
         """
-        return self._levels
+        x = numpy.asarray(x)
```
mphoward, line 411: Why is this needed here, but not in `_function`? I would think both need it?

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+        """list: Domain the sample points come from."""
+        pass
+
+    @property
+    def points(self):
+        """list: Points stored by the set."""
+        return self._points
+
+    @abc.abstractmethod
+    def generate_points(self, num_points):
```
mphoward: This should be a private method taking no arguments that is invoked by `points` when the points need to be accessed
```suggestion
    def _create_points(self):
```
It should return the points.

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+    approximation.
+    The natural domain of the points specified by :param:``natural_domain`` is
+    the domain in which all the possible points in the set lies within.
+    ``number_points`` is the number of points that the UnidimensionalPointSet
+    is expected to store.
+    ``domain`` is the domain of the points generated by the point set.
+    """
+
+    def __init__(self):
+        self._points = []
```
mphoward: Default to not creating them
```suggestion
        self._points = None
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+    @property
+    @abc.abstractmethod
+    def domain(self):
+        """list: Domain the sample points come from."""
+        pass
+
+    @property
+    def points(self):
+        """list: Points stored by the set."""
+        return self._points
```
```suggestion
        if not self._points:
            self._points = self._create_points()
        return self._points
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+    Chebyshev polynomials of the first kind on the domain :math:`[-1, 1]`:
+
+    .. math::
+
+        x_i^* = -\cos(\pi i/n), i = 0,...,n
+
+    For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
+    """
+
+    def __init__(self):
```
mphoward: This constructor needs an argument, like `degree` or `N` representing the degree of the polynomial these points come from.
```suggestion
    def __init__(self, degree):
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+        super().__init__()
+
+    @property
+    def domain(self):
+        """list: Domain the sample points come from."""
+        return [-1, 1]
+
+    def generate_points(self, num_points):
```
```suggestion
    def _create_points(self):
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        if num_points == 0:
+            self._points = []
+        degree = num_points - 1
+        if degree > 0:
+            self._points = list(
+                -numpy.cos(numpy.pi * numpy.linspace(0, degree, degree + 1) / degree)
+            )
+        else:
+            self._points = [0]
```
```suggestion
        if self.degree > 0:
            points = -numpy.cos(numpy.pi * numpy.linspace(0, self.degree, self.degree + 1) / degree)
        else:
            points = numpy.zeros(1)
        return points
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+        return self
+
+
+class NestedClenshawCurtisPointSet(ClenshawCurtisPointSet):
```
mphoward: This class should have its own constructor that takes something like the `max_level`
```py
def __init__(self, max_level):
    self.max_level = max_level
```

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+    intermediary empty levels, but the general order that points are added to
+    levels remains the same.
+
+    This class will generate the nested extrema in the order that corresponds
+    to the exponential growth. For increasing k, the unique extrema of Chebyshev
+    polynomial with degree 2^{k - 1} will be added followed by the extrema of the
+    polynomial with degree 2^{k} , then 2^{k + 1} , then 2^{k + 2} and so on.
+    """
+
+    def generate_points(self, num_points):
```
mphoward: This method should then create the points based on the `max_level`

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
+import abc
+import sys
+import numpy
+
+
+class UnidimensionalPointSet(abc.ABC):
```
mphoward: Should all `UnidimensionalPointSet`s be given a concept of a level the points are assigned at? For sets that don't really have levels, they can all be dumped into the first level.

smolyay/samples.py
```diff
@@ -0,0 +1,207 @@
...
+            else:
+                indexes = numpy.linspace(1, degree - 1, degree - 1, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
+            new_points = list(-numpy.cos(numpy.pi * indexes / degree))
+            points.extend(new_points)
+        self._points = points[:num_points]
+        return self
+
+
+class TrigonometricPointSet(UnidimensionalPointSet):
```
mphoward: Similar changes are needed for these classes as for the Clenshaw-Curtis ones. There also should be a nested and unnested version of the trig points.

mphoward (changes requested):
Thanks Michaela! Here are some comments and requested changes.

smolyay/basis.py
mzbush: That is correct. I'll add it.

smolyay/samples.py
mzbush: Should degree and max_level have getter and setter methods?

smolyay/samples.py
mzbush: With the addition of the `degree` parameter for nonnested point sets and `max_level` for the nested point sets, the Clenshaw-Curtis and Trigonometric sets no longer inherit anything from each other.

Perhaps there should be an abstract class ClenshawCurtis that sets the domain property that both the Clenshaw-Curtis sets can inherit from, and the same thing with the Trigonometric sets.

There could also be a NestedPointSet that is an abstract class for any point set that uses the `max_level` parameter, and a NonNestedPointSet for any point set that uses the `degree` parameter. It's possible for Python classes to have multiple parents, so it would be nice to have the point set objects to clearly demonstrate how they are related via their inheritance structure.

What do you think?

smolyay/samples.py
mzbush: It depends on whether all the MultidimensionPointSets that use a UnidimensionalPointSet will care about levels. A smolyak point set will care, but a tensor point set doesn't have to.

smolyay/samples.py
mphoward: Yes, since we will be caching the values of `points`, you need to add a property setter for any attributes that invalidate the cache if their value changes.

smolyay/samples.py
mphoward: Fair enough. I am good with any of the following options:

1. Force all `UnidimensionalPointSet`s to assign their points to a level (and perhaps default it to `None` if there isn't a concept of a level for a certain set).
2. Create a `NestedUnidimensionalPointSet` that inherits from `UnidimensionalPointSet`. This derived class will add the levels as an abstract concept. The `_create_points` method should then be required to also handle the level assignment. If we go with this option, it would be better for the `_create_points` method to set `self._points` (and `self._levels` or whatever it's called for this class) directly rather than return by value.
3. Allow classes to add an additional `levels` attribute themselves, and whoever consumes the class will use duck typing to accept anything with `points` and `levels.

I'm not so worried about `MultidimensionalPointSet` because these objects can decide whether they want to use the `levels` or not.

smolyay/samples.py
mphoward: `degree` and `max_level` are parameters of the constructors for these concrete classes, but I don't think they should be required as part of the abstract base class interfaces. I probably wouldn't add another base class just to avoid defining the domain property twice.

If you want to use more inheritance, the nested classes should still be able to use the non-nested constructor to figure out all the points by computing the degree from the max level, right? However, then there is still the issue of assigning points to levels, and I would assume that would best be done at the same time as the points are made, so that probably limits the usefulness of this.

smolyay/samples.py
mzbush: I've been hesitant to add level information to the UnidimensionalPointSets and stuck with just number of points at first because if the nested point sets are dependent on the number of levels to determine how many points to generate, then that makes it more cumbersome to use growth schemes that aren't the default. Slow growth and every other growth scheme would need its own class for every unique set of points. I was considering having the levels handled by the MultidimensionalPointSets or by some other object that would exclusively handle levels.

If NestedUnidimensionalPointSets are expected to handle levels, should there be another variable that contains information about how the set is expected to grow each level? The paper I linked in the surrogate models slack channel gives the order function for a ton of growth schemes, which gives exactly how many total points the levels have accumulated. An `order` parameter that's either a callable or a list could be taken with the `max_level` parameter to determine what points are needed and what the list of level indexes should look like. Alternatively, `order` could be a variable that allows the user to select from a predetermined set of allowed order functions, which is what the sgmga code does.

mphoward: To me, how the levels for points are determined is a detail that deriving classes can figure out. This could be achieved in several ways, so I think for now, we just need to decide whether the base unidimensional class has the concept of a level, or if that only belongs in the nested version. Once we make that decision, we would need to survey all the ways for assigning points to levels that we reasonably think we need to support and make a decision about how best to do that.

smolyay/samples.py
mzbush: For the Trigonometric point set, is degree the most appropriate name for that variable? While `degree` makes sense for the Clenshaw Curtis point set since it's the degree of the Chebyshev polynomial, what is the "degree" in the context of the Trigonometric point set?

mphoward: No, probably not. You can choose an appropriate name and use it in that class. It does not need to be the same as the one used in Clenshaw-Curtis points

mzbush: I would suggest changing it to number of points, so `numpoint` or `number_points`, but that's what degree was used to replace in the first place and I'd want the nonnested point sets to have parameters consistent with each other.

mphoward: I don't think number of points makes sense for the trigonometric points because they should be the roots of a certain frequency trig function. So, maybe `frequency` makes sense here?

And just to emphasize again: specific constructors for derived classes do not need to be the same. You should choose terminology that makes sense based on what they represent.

mzbush: Derived classes don't need to be the same as each other, but I feel it's prudent to point out that some amount of consistency between them is more convenient. When these classes are used by a multidim point set, like the one for creating tensor grids, that class isn't going to know what kind of UnidimensionalPointSet it's using nor should it. It shouldn't have to run different code depending on what child class it's using, so there should be some level of consistency.

And on a wider scale, the code using and initializing the `UnidimensionalPointSet` to make grids shouldn't have to care that much about what kind of `UnidimensionalPointSet` it's using either, which makes it easier to reuse code and predict how changes will affect the grids it generates.

As it stands now, the `ClenshawCurtisPointSet` when initialized will create a list of points with 1 more element than the number it was initialized with. The `TrigonometricPointSet` when initialized will create a list of points with as many elements as the number it was initialized with (with 0 being the exception). A person unfamiliar with this code and its inner workings may find this confusing and not understand why their tensor grids are generating with more or less points than they expect. You can't just switch the class type and expect to get a tensor grid with the exact same number of points unless you remember to change the input. 

And I imagine that would be very frustrating, especially since it would be trivial to make both classes give the same number of points as the input parameter.

It's the whole reason why `number_points` was a property of the base class when I first made the pull request. Even if you knew nothing about the deriving classes' inner workings, you would always know how many points you would get.

smolyay/samples.py
mzbush: I think that levels should only be a concept for the point sets that the smolyak multidim point set is expected to be able to use. This excludes `ClenshawCurtisPointSet` and `TrigonometricPointSet`, so the base class shouldn't have a concept of levels.

I propose a base class called `LevelUnidimensionalPointSet` or `TieredUnidimensionalPointSet`. (Not `NestedUnidimensionalPointSet` because while the point sets we're using for smolyak are nested, the sets it can use don't necessarily have to be nested.) It will be a child of `UnidimensionalPointSet` and parent to all classes that use levels and that the smolyak grid generator will be allowed to use. 

It will take a parameter `max_level`, and it will have additional read-only properties `levels` and `growth_order`. The `levels` property is identical to the `levels` in the `NestedBasisFunctionSet`, and the base class will have a method for calculating it from `growth_order` and `max_level`. The `growth_order` is an abstract property like `domain` and it is either a callable or a list. It outputs the expected accumulated number of points at each level. Children of this class are expected to give concrete versions of `domain`, `growth_order`, and `_create_points`.

(The reason that `growth_order` is allowed to be callable or a list is to accommodate finite point sets.)

`NestedClenshawCurtisPointSet` would set the `growth_order` to be `lambda x : 1 if x==0 else 2**x + 1`. Replacing its code for generating points with the code that I sent in slack yesterday, I would be able to derive children of `NestedClenshawCurtisPointSet` for slow growth, moderate growth, and linear 2 growth that only need to override `growth_order`.

smolyay/samples.py
mphoward: > Derived classes don't need to be the same as each other, but I feel it's prudent to point out that some amount of consistency between them is more convenient. ... And on a wider scale, the code using and initializing the UnidimensionalPointSet to make grids shouldn't have to care that much about what kind of UnidimensionalPointSet it's using either

Yes, this is the whole idea of duck typing and defining an abstract base class with a common interface. The minimum attributes and methods that all concrete implementations need should be in the abstract base class, but this **does not** mean that their constructors have to be the same or that deriving classes cannot have attributes that are not in the base class. This is a fundamental concept in object-oriented program.

> You can't just switch the class type and expect to get a tensor grid with the exact same number of points unless you remember to change the input.

Of course, but there is no reason we need to do this.

> Even if you knew nothing about the deriving classes' inner workings, you would always know how many points you would get.

I can get the number of points for any point set by doing this...

```py
num_points = len(point_set.points)
```

smolyay/samples.py
mphoward: > I propose a base class called LevelUnidimensionalPointSet or TieredUnidimensionalPointSet. (Not NestedUnidimensionalPointSet because while the point sets we're using for smolyak are nested, the sets it can use don't necessarily have to be nested.)

I like this idea and agree with not using the term nested!

> It will take a parameter max_level, and it will have additional read-only properties levels and growth_order. ... The growth_order is an abstract property like domain and it is either a callable or a list. It outputs the expected accumulated number of points at each level.

Doing this will inherently make things nested though, won't it? Since it would not be possible for a point to belong to multiple levels, or to have points added in a non-nested fashion.

I propose that you make a base class that has the levels property, but not the other stuff. I also propose that the `_create_points` method be renamed to `_create` in the `UnidimensionalPointSet` class. For normal point sets, it is expected to assign the value of `self._points`. For your level class, it should assign values of both `self._points` and `self._levels`. The `points` (and `levels`) properties should then invoke this method when an internal cache variable has been invalidated. You can put that `_valid_cache` bool in the `UnidimensionalPointSet` base class, then check it as needed rather than checking if `self._points is None`. You set it to `False` in the setters that invalidate the cache.

We can then think about the mechanism for assigning points to levels in the `_create` method of deriving classes.

smolyay/samples.py
mzbush: > I can get the number of points for any point set by doing this...
> num_points = len(point_set.points)

Yes, but you can't ask the `UnidimensionalPointSet` for a set with a desired number of points. You have to convert the number of points you want into a number corresponding to whatever parameter the class actually asks for. 

Having `UnidimensionalPointSet` ask for a number of points as a parameter is what I would prefer.

`ClenshawCurtisPointSet` does the same thing as `numpy.polynomial.chebyshev.chebpts2`, with the key difference being that `ClenshawCurtisPointSet` will output [0] while `chebpts2` will raise an error if you ask for less than 2 points. That function asks for number of points, and I can easily see someone new to this code expecting `ClenshawCurtisPointSet` to do the same thing.

What I don't understand is, why do you think the degree of the polynomial the points are the extrema of is so important that it should be the parameter the class asks for?

mphoward: Because doing so hides the math related to what level of approximation the points are derived from. It is straightforward to convert the degree of the Chebyshev polynomial or frequency of a trig function to a number of points, but going the other way around loses guarantees on having a complete set of points associated with a certain level of approximation. What I'm describing is also how these points would typically be constructed in 1d interpolation.

mphoward: > you can't ask the UnidimensionalPointSet for a set with a desired number of points

What is the use case you have in mind where this would be needed?

smolyay/samples.py
mzbush: > Doing this will inherently make things nested though, won't it? Since it would not be possible for a point to belong to multiple levels, or to have points added in a non-nested fashion.

In a fully nested rule, the `growth_order` gives the accumulated unique points for a given and all its previous levels, so total new points added at level L is growth_rule(L) - growth_rule(L-1). For a weakly nested (0.0 recurs) or nonnested rule (nothing recurs) this isn't the case, but if a deriving class is nonnested or weakly nested, it can override the method for generating levels.

For nonnested rules it appears to work such that 

`new_points_added(L) = growth_rule(L) if growth_rule(L) - growth_rule(L - 1) != 0 else 0`

and weakly nested rules do this 

`new_points_added(L) = growth_rule(L)-1 if growth_rule(L) - growth_rule(L - 1) != 0 else 0`

For very weakly nested rules, it depends on what levels 0.0 recurs.

The `growth_order` given a max level can easily calculate how many total points are needed, and the number of points can be sent to `_create_points` to generate that many points.

> I also propose that the `_create_points` method be renamed to `_create` in the `UnidimensionalPointSet` class. For normal point sets, it is expected to assign the value of `self._points`. For your level class, it should assign values of both `self._points` and `self._levels`. The `points` (and `levels`) properties should then invoke this method when an internal cache variable has been invalidated.

The internal cache and having the abstract method `_create` for assigning all relevant variables is fine, but I still feel that `points` and `levels` should be handled separately if possible, and not in the same for loop like the old code. Forcing them to both be computed in the same subroutine means that any time we want to use different points or a different growth scheme the entire method for generating both has to be rewritten. Doing them separately isn't going to result in massive runtime penalties. TieredUnidimensionalPointSet can have a defined `_create` method that calls methods `_create_points` and `_create_levels`. Having the `growth_order` property makes creating both as well as the process of generating new classes based off old ones easier.

smolyay/samples.py
mzbush: > What I'm describing is also how these points would typically be constructed in 1d interpolation.

I looked up how scipy does it with the roots of the other quadratures, and I will concede that you are right. `ClenshawCurtisPointSet`will use degree.

(Interestingly, they all have the same number of roots as the degree of the polynomial. This explains why Clenshaw Curtis quadrate is the only one where the exponential growth is described by `2**x + 1` in the "1D Quadrature Rules for Sparse Grids" paper, while the rest use `2**(x+1) - 1`.)

smolyay/samples.py
mphoward: > Forcing them to both be computed in the same subroutine means that any time we want to use different points or a different growth scheme the entire method for generating both has to be rewritten.

This is not necessarily the case. There are ways to package the code for assigning the levels in a reusable way, but we need to talk about what cases we want to enable first.

> Having the growth_order property makes creating both as well as the process of generating new classes based off old ones easier.

I don't think this is necessarily the case either.

smolyay/basis.py
```diff
@@ -87,24 +127,14 @@ class ChebyshevFirstKind(BasisFunction):
 
     def __init__(self, n):
         super().__init__()
-        self._n = n
-        self._derivative_polynomial = None
-        if n > 0:
-            self._points = [-numpy.cos(numpy.pi*i/n) for i in range(n+1)]
-        else:
-            self._points = [0]
+        self.n = n
```
mphoward: You might want to do this to force a copy and also make sure we have an int
```suggestion
        self.n = int(n)
```
Sorry to backtrack on this, but this actually means that `n` should probably be a property with a getter and a setter. The setter should do this int cast.

smolyay/basis.py
```diff
@@ -328,141 +278,138 @@ def derivative(self, x):
...
-        """list: List of lists of indexes for points/functions at each level.
         Raises
         ------
-        IndexError
-            max index must be less than total number of points.
+        ValueError
+            If input is outside the domain `[0, 2\pi]`
+        """
+        x = numpy.asarray(x)
+        return numpy.exp(x * self.sigma * 1j)
```
mphoward: I would expect this basis function to use `n` here (positive or negative) rather than `sigma`. I think `sigma` is more like an internal variable that would be used to nest these. Please confirm with Mohammadreza, and then probably remove `sigma` from this class

smolyay/basis.py
```diff
@@ -251,24 +213,14 @@ class ChebyshevSecondKind(BasisFunction):
...
-        if n > 1:
-            self._points = [-numpy.cos(k*numpy.pi/(n+1)) for k in range(1, n+1)]
-        else:
-            self._points = [0.]
-
-    @property
-    def points(self):
-        """list: Sampling points at extrema of polynomial."""
-        return self._points
+        self.n = n
```
mphoward: Same comments on `n`.

smolyay/basis.py
```diff
@@ -328,141 +278,138 @@ def derivative(self, x):
...
 
-    def __init__(self,points,basis_functions):
-        self._basis_functions = basis_functions
-        self._points = points
-        if len(basis_functions) != len(points):
-            raise IndexError("basis_functions and points must have the "
-                    "same number of elements.")
+    def __init__(self, n):
+        super().__init__()
+        self.n = n
```
mphoward: Same comments on `n`

tests/basis_test.py
```diff
@@ -3,52 +3,21 @@
...
-def test_cheb_initial_8():
-    """test initial when degree is 8"""
-    expected_points = [-1, -numpy.sqrt(numpy.sqrt(2)+1)/(2**0.75),
-            -1/numpy.sqrt(2), -numpy.sqrt(numpy.sqrt(2)-1)/(2**0.75), 0,
-            numpy.sqrt(numpy.sqrt(2)-1)/(2**0.75), 1/numpy.sqrt(2),
-            numpy.sqrt(numpy.sqrt(2)+1)/(2**0.75), 1]
-    f = ChebyshevFirstKind(8)
-    assert f.n == 8
-    assert numpy.allclose(f.points,expected_points,atol=1e-10)
+from smolyay.basis import (
```
mphoward: I would rather see these used in their full namespace rather than using the `from ... import`. So, `smolyay.basis.ChebyshevFirstKind` for example.

tests/basis_test.py
```diff
@@ -58,14 +27,16 @@ def test_cheb_call_degree_0_1():
         assert f0(i) == 1
         assert f1(i) == i
 
+
 def test_cheb_call_random_points():
```
mphoward: In addition to checking random points, might you also add a test where you check special points, like zeros or extrema, that have known values?

tests/basis_test.py
```diff
@@ -74,9 +45,9 @@ def test_cheb_call_random_points_multi_input():
     f0 = ChebyshevFirstKind(0)
```
mphoward: These sorts of variations of values should be done using a parametrization rather than repeating the code 3 times.

tests/basis_test.py
```diff
@@ -58,14 +27,16 @@ def test_cheb_call_degree_0_1():
         assert f0(i) == 1
         assert f1(i) == i
 
+
 def test_cheb_call_random_points():
     """Test chebyshev polynomial at some degree at some input"""
     numpy.random.seed(567)
-    ns = numpy.random.randint(20,size = 20)
+    ns = numpy.random.randint(20, size=20)
```
mphoward: This test should be parametrized on a few n, rather than generating a random list of n within the test. You won't need the for loop then.

tests/basis_test.py
```diff
@@ -74,9 +45,9 @@ def test_cheb_call_random_points_multi_input():
...
     fn = ChebyshevFirstKind(5)
-    assert numpy.allclose(f0(xs),special.eval_chebyt(0,xs))
-    assert numpy.allclose(f1(xs),special.eval_chebyt(1,xs))
-    assert numpy.allclose(fn(xs),special.eval_chebyt(5,xs))
+    assert numpy.allclose(f0(xs), special.eval_chebyt(0, xs))
+    assert numpy.allclose(f1(xs), special.eval_chebyt(1, xs))
+    assert numpy.allclose(fn(xs), special.eval_chebyt(5, xs))
 
 
 def test_cheb_derivative():
```
mphoward: Please apply the comments about parametrizing the tests everywhere it is relevant in this file. You can pair n's with expected values, if you know them.

mphoward (changes requested):
Thanks Michaela! Here are some comments on the code and tests. Overall, this is in good shape.

smolyay/basis.py
mzbush: He said it's okay to remove it.

smolyay/basis.py
```diff
@@ -87,24 +127,23 @@ class ChebyshevFirstKind(BasisFunction):
 
     def __init__(self, n):
         super().__init__()
-        self._n = n
-        self._derivative_polynomial = None
-        if n > 0:
-            self._points = [-numpy.cos(numpy.pi*i/n) for i in range(n+1)]
-        else:
-            self._points = [0]
+        self._n = int(n)
```
mphoward: Do like this so it goes through the setter
```suggestion
        self.n = int(n)
```

smolyay/basis.py
```diff
@@ -87,24 +127,23 @@ class ChebyshevFirstKind(BasisFunction):
 
     def __init__(self, n):
```
mphoward: One final comment here: Python usually encourages human-readable variable names rather than single letters. Should we call this `degree` rather than `n` throughout the two Chebyshev classes?

smolyay/basis.py
```diff
@@ -328,141 +296,151 @@ def derivative(self, x):
...
 
-    @classmethod
-    def make_nested_set(cls, exactness):
-        """Create a nested set of Chebyshev polynomials.
 
-        A nested set is created up to a given level of ``exactness``,
-        which corresponds to a highest-order Chebyshev polynomial of
-        degree ``n = 2**(exactness + 1) - 1``.
+class Trigonometric(BasisFunction):
+    r"""Trigonometric basis functions.
```
mphoward, line 328: Please update the documentation for this class, it is out of date after the changes made to remove sigma.

Also, what would be the appropriate new name for `n`? It could also be `degree`, if you add some documentation explaining that:
```
e^{i n x} = (e^{ix})^n
```
It could also be called `frequency` or something like that, but then there is the weirdness of it being allowed to be a negative number (which frequencies usually are not).

Maybe consult with @mzf0069 on these changes

smolyay/basis.py
```diff
@@ -328,141 +296,151 @@ def derivative(self, x):
...
+        Raises
+        ------
+        ValueError
+            If input is outside the domain `[0, 2\pi]`
+        """
+        x = numpy.asarray(x)
+        if self.n % 2 == 1:
+            return numpy.exp(x * (1 + self.n) / 2 * 1j)
+        else:
+            return numpy.exp(x * -self.n / 2 * 1j)
```
mphoward: I think there may have been a miscommunication before. This should just be `n` and `n` is allowed to be negative:
```suggestion
        return numpy.exp(x * self.n  * 1j)
```

smolyay/basis.py
```diff
@@ -328,141 +296,151 @@ def derivative(self, x):
...
-            max index must be less than total number of points.
+        ValueError
+            If input is outside the domain `[0, 2\pi]`
         """
-        return self._levels
+        x = numpy.asarray(x)
+        if self.n % 2 == 1:
+            return (1 + self.n) / 2 * 1j * numpy.exp(x * (1 + self.n) / 2 * 1j)
+        else:
+            return -self.n / 2 * 1j * numpy.exp(x * -self.n / 2 * 1j)
```
mphoward:
```suggestion
        return self.n * 1j * numpy.exp(x * self.n * 1j)
```

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
-
-def test_cheb_call_random_points():
-    """Test chebyshev polynomial at some degree at some input"""
-    numpy.random.seed(567)
-    ns = numpy.random.randint(20,size = 20)
-    xs = numpy.random.rand(20) * 2 - 1
-    for n,x in zip(ns,xs):
-        f = ChebyshevFirstKind(n)
-        assert numpy.isclose(f(x),special.eval_chebyt(n,x))
+import smolyay.basis
```
mphoward: Just import `smolyay`
```suggestion
import smolyay
```
If it isn't working like that, then we should fix the import structure of the package. Let me know

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
-    assert numpy.isclose(u0.derivative([1,-1,0.5]),[0,0,0]).all()
-    assert numpy.isclose(u1.derivative([1,-1,0.5]),[2,2,2]).all()
-    assert numpy.isclose(u2.derivative([1,-1,0.5]),[8,-8,4]).all()
 
+def test_cheb_n_setter():
+    """test degree setter"""
+    f2 = smolyay.basis.ChebyshevFirstKind(2)
+    f2.n = 3
+    assert f2.n == 3
+    assert isinstance(f2.n, int)
```
mphoward: You can make these one test, no need to do separately

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+
+
+@pytest.mark.parametrize("n", list(range(20)))
```
mphoward: Is it really necessary to test this for 20 different values of n? I would test it for 0, 1, and 2 since the first are special cases but the last is representative of all n.

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+def test_cheb_call_random_points(n):
+    """Test chebyshev polynomial at some degree at some input"""
+    numpy.random.seed(567)
+    xs = numpy.random.rand(20) * 2 - 1
```
mphoward: What is the benefit of doing this randomly instead of just linearly?
```suggestion
    xs = numpy.linspace(-1, 1, 20)
```

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+    f = smolyay.basis.ChebyshevFirstKind(n)
+    assert numpy.allclose(f(xs), special.eval_chebyt(n, xs))
+
+
+def test_cheb_call_extrema_points():
+    """Test chebyshev polynomial at some degree at some input"""
+    extrema_points = [-1.0, -1 / numpy.sqrt(2), 0, 1 / numpy.sqrt(2), 1]
+    extrema_output = [1, -1, 1, -1, 1]
+    f = smolyay.basis.ChebyshevFirstKind(4)
```
mphoward: It is easier to read this with the function coming first, since then we know why these are the extrema.
```suggestion
    f = smolyay.basis.ChebyshevFirstKind(4)
    extrema_points = [-1.0, -1 / numpy.sqrt(2), 0, 1 / numpy.sqrt(2), 1]
    extrema_output = [1, -1, 1, -1, 1]
```

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+def test_cheb_call_root_points():
+    """Test chebyshev polynomial at some degree at some input"""
+    root_points = [
+        -numpy.sqrt(numpy.sqrt(2) + 1) / (2**0.75),
+        -numpy.sqrt(numpy.sqrt(2) - 1) / (2**0.75),
+        numpy.sqrt(numpy.sqrt(2) - 1) / (2**0.75),
+        numpy.sqrt(numpy.sqrt(2) + 1) / (2**0.75),
+    ]
+    f = smolyay.basis.ChebyshevFirstKind(4)
+    assert numpy.allclose(f(root_points), [0, 0, 0, 0])
```
mphoward:
```suggestion
    f = smolyay.basis.ChebyshevFirstKind(4)
    root_points = [
        -numpy.sqrt(numpy.sqrt(2) + 1) / (2**0.75),
        -numpy.sqrt(numpy.sqrt(2) - 1) / (2**0.75),
        numpy.sqrt(numpy.sqrt(2) - 1) / (2**0.75),
        numpy.sqrt(numpy.sqrt(2) + 1) / (2**0.75),
    ]
    assert numpy.allclose(f(root_points), numpy.zeros(4))
```

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+
+
+@pytest.mark.parametrize("n", list(range(20)))
```
mphoward: Move this test up to come after the other one that tests evaluation of a list. You should also apply my other comments about that test here.

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
+def test_cheb_call_random_points_multi_input(n):
+    """Test chebyshev polynomial call handles inputs with complex shape"""
+    numpy.random.seed(567)
+    xs = numpy.random.rand(20, 3, 4, 6) * 2 - 1
+    f = smolyay.basis.ChebyshevFirstKind(n)
+    answer = f(xs)
+    answer_check = special.eval_chebyt(n, xs)
+    assert answer.shape == answer_check.shape
+    assert numpy.allclose(answer, answer_check)
+
```
mphoward: I don't see a test for a single input? You should adda test to check that and make sure it comes back as the right float (not a numpy array).

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
 
 def test_cheb_derivative_invalid_input():
-    """Test call raises error if input is outside domain [-1,1]"""
-    f = ChebyshevFirstKind(4)
+    """Test call raises error if input is outside domain [-1, 1]"""
+    f = smolyay.basis.ChebyshevFirstKind(4)
     with pytest.raises(ValueError):
         f.derivative(2)
     with pytest.raises(ValueError):
         f.derivative(-2)
```
mphoward: Add a check on a list and list of lists here too?

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
-    f = ChebyshevFirstKind(4)
+    """Test call raises error if input is outside domain [-1, 1]"""
+    f = smolyay.basis.ChebyshevFirstKind(4)
     with pytest.raises(ValueError):
         f(2)
     with pytest.raises(ValueError):
         f(-2)
     with pytest.raises(ValueError):
-        f([0.5,0.7,3,0.8])
+        f([0.5, 0.7, 3, 0.8])
```
mphoward: Add a test for a list of lists here too?

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
-def test_cheb_2nd_call_degree_0_1():
-    """Chebyshev polynomial degree 0 is always 1 and degree 1 returns 2*input"""
-    f0 = ChebyshevSecondKind(0)
-    f1 = ChebyshevSecondKind(1)
-    for i in [-1, -0.5, 0, 0.5, 1]:
-        assert f0(i) == 1
-        assert f1(i) == i*2
-
-def test_cheb_2nd_call_random_points():
+def test_cheb_2nd_initial():
```
mphoward: Please apply my comments on `ChebyshevFirstKind` forward to the other basis functions

tests/basis_test.py
```diff
@@ -3,328 +3,296 @@
...
-def test_make_nested_empty_cheb_2nd():
-    """Check make_nested_set makes empty NestedBasisFunctionSet"""
-    f = ChebyshevSecondKind.make_nested_set(0)
-    assert f.points == [0,0]
-    assert f.levels == [[0,1]]
-    basis_functions = f.basis_functions
-    assert len(basis_functions) == 2
-    assert basis_functions[0].n == 0
 
+def test_set_initialize_0():
```
mphoward: Can you merge this with the other basis function set test? Given that this class has been pared down and is not the focus of this PR, having this one lightweight test is sufficient for now. We'll revisit again in future.

mphoward (changes requested):
Thanks! I have some comments, mostly on the tests, but this is close to done.

tests/basis_test.py
mzbush: Another comment implied that the for loop in the call_random_points tests should be removed, and part of its purpose was to do single input checks.

smolyay/basis.py
mzbush: The reason it's called n is because that's the symbol used in the mathematical notation and what is used by scipy. `deg` is used by numpy, and `degree` is what's used by the point set. I would prefer to use naming conventions consistent with one of these packages, so I'll use what we have for the point set.

tests/basis_test.py
mzbush: It doesn't work.

tests/basis_test.py
mzbush: My testing philosophy is to test slightly more than you think you have to, so you test all the special cases and multiple representative cases if multiple representative cases is fairly trivial to do. It would, after all, be a shame if something went wrong for an `n` one higher than you tested for. Here it's really easy to test for n higher than 2 so there's no reason not to.

I can do 5 instead to reduce the sheer number of tests while covering more than just the first trivial non-special case.

smolyay/basis.py
mzbush: I would prefer `n`, whatever it represents, to correspond with the parameter used by the TrigonometricPointSet, but I want @mzf0069 to weigh in before doing anything or updating the documentation.

mphoward: They do not necessarily need to be the same because they are different objects

tests/basis_test.py
mphoward: These tests will need to be run in multiple build configurations on every commit, so you should not run more than you actually need in order to keep the required time reasonable. We can add additional test cases if bugs are ever reported. Please restrict to 0, 1, and 2.

tests/basis_test.py
mphoward: We must be missing imports in `__init__.py`. In that file, add:
```py
from . import basis
```
You should do this for all module files to bind them to the package namespace. See here for more information:

https://docs.python.org/3/reference/import.html#package-relative-imports

smolyay/basis.py
mphoward: `degree` is more descriptive than `n` or `deg`. We will probably also enable a code linter that will reject short variable names like this, so please use this. In the class documentation, you can still write something like:
```py
the `degree` *n*
```
and then use *n* in any mathematical formulas.

tests/basis_test.py
mphoward: Yes, the for loop should have been removed, but you should also still do a single input test. Just write a test that calls the function for a value, say 0.5.

smolyay/basis.py
mzf0069: > Please update the documentation for this class, it is out of date after the changes made to remove sigma.
> 
> Also, what would be the appropriate new name for `n`? It could also be `degree`, if you add some documentation explaining that:
> ```
> e^{i n x} = (e^{ix})^n
> ```
> It could also be called `frequency` or something like that, but then there is the weirdness of it being allowed to be a negative number (which frequencies usually are not).
> 
> Maybe consult with @mzf0069 on these changes

I would rename it to `node` or `node_frequency`. Each frequency has two nodes (one negative and the other positive --- except the first one).

mphoward: Hmm, interesting. Including `node` works nicely for interpolation, but this is a more general concept of a basis function. I would use just `frequency`. Negative frequencies are fine mathematically:
https://en.wikipedia.org/wiki/Negative_frequency

mphoward: Or `wavenumber`?

mzf0069: > Hmm, interesting. Including `node` works nicely for interpolation, but this is a more general concept of a basis function. I would use just `frequency`. Negative frequencies are fine mathematically:
> https://en.wikipedia.org/wiki/Negative_frequency

If negative frequency are fine, I would probably stick with `frequency` since the phrase is more common in the trig sparse grid papers.

mzbush: Okay, I've made the changes. Frequency can be any integer and I've adjusted the call and the tests to accommodate the negative frequencies.

tests/basis_test.py
mzbush: done!

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
...
+import smolyay
+
+
+def test_cheb_initial():
+    """test degrees returns correctly"""
+    f2 = smolyay.basis.ChebyshevFirstKind(2)
+    assert f2.degree == 2
+    assert isinstance(f2.degree, int)
+    assert numpy.array_equal(f2.domain, [-1, 1])
+    f2.degree = 3
```
mphoward: Maybe you should assign this as a float to make sure it is silently converted to an `int`?
```suggestion
    f2.degree = float(3)
```

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
...
+    assert f2.degree == 3
+    assert isinstance(f2.degree, int)
+
+
+@pytest.mark.parametrize("degree", list(range(3)))
```
mphoward: This is more readable:
```suggestion
@pytest.mark.parametrize("degree", [0, 1, 2])
```

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
 import numpy
 from scipy import special
```
mphoward: Try to avoid using from imports, as they pollute the namespace
```suggestion
import scipy.special
```
Update as needed below

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
...
-
-def test_cheb_2nd_derivative():
-    """Test if the correct derivative is generated."""
-    u0 = ChebyshevSecondKind(0)
-    u1 = ChebyshevSecondKind(1)
-    u2 = ChebyshevSecondKind(2)
-    assert u0.derivative(1) == pytest.approx(0)
-    assert u1.derivative(1) == pytest.approx(2)
-    assert u2.derivative(1) == pytest.approx(8)
+@pytest.mark.parametrize("degree", list(range(3)))
```
mphoward:
```suggestion
@pytest.mark.parametrize("degree", [0, 1, 2])
```

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
...
+def test_cheb_call_random_points_multi_input(degree):
+    """Test chebyshev polynomial call handles inputs with complex shape"""
+    f = smolyay.basis.ChebyshevFirstKind(degree)
+    numpy.random.seed(567)
+    xs = numpy.random.rand(20, 3, 4, 6) * 2 - 1
```
mphoward: It would be better not to use random points. You also you don't need so many points along each dimension, but you do want to have more systematic coverage of certain dimension sizes.

I would update this test so that it covers first 2d arrays, including special cases where the array shape is 1 along one or both dimensions.

You can then write a test that covers 3d arrays, including these special cases again. There is probably no need to go to higher dimensionality than that.

You also need a test that covers 1d arrays. It seems like that might have gotten lost when you added the tests on single inputs? That should cover similarly the case where the array has only 1 element and where it has more than one element.

It might be better to make each of these its own test function (1d, 2d, and 3d).

tests/basis_test.py
```diff
@@ -3,328 +3,274 @@
...
+
+@pytest.mark.parametrize(
+    "degree,answer",
+    [
+        (0, [0, 0]),
+        (1, [1, 1]),
+        (2, [4, -2]),
+    ],
+)
+def test_cheb_derivative(degree, answer):
```
mphoward: We need coverage of the derivative for the different types of inputs above as well.

mphoward (changes requested):
Thanks! Here are some more comments on the tests. Please make sure to apply the changes that you make to the ChebyshevFirstKind forward to the other basis functions.

tests/basis_test.py
mzbush: Should it also test that the shape is maintained? Like if the input's shape is (1, 5) or (5, 1) or (1, 1, 5, 1, 1, 1)

mphoward: Yes, definitely. The output shape should match the input.

mzbush: Done. There are now tests for scalars, 1D arrays, 2D arrays, and 3D arrays for the call and derivative of each basis function.

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+        (2, {0.5: -0.5, 1: 1, -1: 1, -0.25: -0.875, -0.5: -0.5}),
+    ],
+)
+def test_cheb_call_1D(degree, answer_key):
+    """Test chebyshev polynomial call with a 1D array"""
+    f = smolyay.basis.ChebyshevFirstKind(degree)
+    get_answer = lambda x: numpy.select(
+        [numpy.array(x) == k for k in answer_key], answer_key.values(), numpy.nan
+    )
+    xs = [0.5, 1, -1, -0.25, -0.5]
```
mphoward: Will this work instead?
```suggestion
    xs = [0.5, 1, -1, -0.25, -0.5]
    answers = [answer_key[x] for x in xs]
```

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+    [
+        (0, {0.5: 1, 1: 1, -1: 1, -0.25: 1, -0.5: 1}),
+        (1, {0.5: 0.5, 1: 1, -1: -1, -0.25: -0.25, -0.5: -0.5}),
+        (2, {0.5: -0.5, 1: 1, -1: 1, -0.25: -0.875, -0.5: -0.5}),
+    ],
+)
+def test_cheb_call_2D(degree, answer_key):
+    """Test chebyshev polynomial call with a 2D array"""
+    f = smolyay.basis.ChebyshevFirstKind(degree)
+    get_answer = lambda x: numpy.select(
```
mphoward: I think you can do the same thing I suggested above here, but reshape the answers as well. The easiest way to do this would be to make xs be a 1d array, answers a 1d array, and then reshape both into the various 2d shapes

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+        [numpy.array(x) == k for k in answer_key], answer_key.values(), numpy.nan
+    )
+    xs = [[0.5, 1, -1, -0.25], [-0.5, -1, 1, -0.25]]
+    xs2 = numpy.reshape(xs[0], (1, 4))
+    xs3 = numpy.reshape(xs[0], (4, 1))
```
mphoward: Can we get a test where the shape is `(1,1)` too?

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+)
+def test_cheb_call_1D(degree, answer_key):
+    """Test chebyshev polynomial call with a 1D array"""
+    f = smolyay.basis.ChebyshevFirstKind(degree)
+    get_answer = lambda x: numpy.select(
+        [numpy.array(x) == k for k in answer_key], answer_key.values(), numpy.nan
+    )
+    xs = [0.5, 1, -1, -0.25, -0.5]
+    assert numpy.shape(f(xs)) == numpy.shape(xs)
+    assert numpy.allclose(f(xs), get_answer(xs))
```
mphoward: Can we also get a test where the shape is `(1,)`?

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+    f = smolyay.basis.ChebyshevFirstKind(degree)
+    get_answer = lambda x: numpy.select(
+        [numpy.array(x) == k for k in answer_key], answer_key.values(), numpy.nan
+    )
+    xs = [[0.5, 1, -1, -0.25], [-0.5, -1, 1, -0.25]]
+    xs2 = numpy.reshape(xs[0], (1, 4))
+    xs3 = numpy.reshape(xs[0], (4, 1))
+    assert numpy.shape(f(xs)) == numpy.shape(xs)
+    assert numpy.allclose(f(xs), get_answer(xs))
+    assert numpy.shape(f(xs2)) == (1, 4)
```
mphoward: Could you please interleave the creation of the various xs and the tests for them, with white space in between? It makes it easier to read if you see the input get created and then immediately tested.

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
+
+@pytest.mark.parametrize(
+    "degree,answer_key",
+    [
+        (0, {0.5: 0, 1: 0, -1: 0, -0.25: 0, -0.5: 0}),
+        (1, {0.5: 1, 1: 1, -1: 1, -0.25: 1, -0.5: 1}),
+        (2, {0.5: 2, 1: 4, -1: -4, -0.25: -1, -0.5: -2}),
+    ],
+)
+def test_cheb_derivative(degree, answer_key):
```
mphoward: Same comments are for call apply to the derivative tests.

tests/basis_test.py
```diff
@@ -1,330 +1,1017 @@
...
-    f1 = ChebyshevSecondKind(1)
-    fn = ChebyshevSecondKind(5)
-    assert numpy.allclose(f0(xs),special.eval_chebyu(0,xs))
-    assert numpy.allclose(f1(xs),special.eval_chebyu(1,xs))
-    assert numpy.allclose(fn(xs),special.eval_chebyu(5,xs))
+    with pytest.raises(ValueError):
+        f.derivative([[0.5, 0.7, 3, 0.8], [0.5, 0.7, 0.8, -0.5]])
+
+
+def test_cheb_2nd_initial():
```
mphoward: Are the ChebyshevSecondKind tests all basically the same as first kind? If yes, then you could make the first parameter of the tests be a basic function rather than a degree to reduce duplication!

mphoward (changes requested):
This looks great. I have a few more requests on the tests, but then I think this will be ready to merge.

tests/basis_test.py
mzbush: done.

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+    assert numpy.array_equal(f2.domain, [-1, 1])
+    f2.degree = float(3)
+    assert f2.degree == 3
+    assert isinstance(f2.degree, int)
+
+
+@pytest.mark.parametrize(
+    "cheb_function,degree,answer_key",
+    [
+        (smolyay.basis.ChebyshevFirstKind, 0, {0.5: 1, 1: 1, -1: 1, -0.25: 1, -0.5: 1}),
```
mphoward: Would this work, with just two parameters `function,answer_key`?
```suggestion
        (smolyay.basis.ChebyshevFirstKind(0), {0.5: 1, 1: 1, -1: 1, -0.25: 1, -0.5: 1}),
```
Seems like that would be easier to read

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+    ],
+)
+def test_cheb_call(cheb_function, degree, answer_key):
+    """Test chebyshev polynomial call"""
+    f = cheb_function(degree)
+    assert f(0.5) == pytest.approx(answer_key[0.5])
+    assert f(1) == pytest.approx(answer_key[1])
+    assert f(-1) == pytest.approx(answer_key[-1])
+    assert f(-0.25) == pytest.approx(answer_key[-0.25])
+    assert f(-0.5) == pytest.approx(answer_key[-0.5])
```
mphoward: I'm going to suggest something that goes against my usual advice here. If we turn this into a loop like:
```suggestion
    for x, y in answer_key.items():
        assert f(x) == pytest.approx(y)
```
and you also make the change I suggested above, can we also just add the `Trigonometric` functions to the parameter list for all of these call tests? That would cut down on a lot of code that's copied below.

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+            smolyay.basis.ChebyshevSecondKind,
+            2,
+            {0.5: 0, 1: 3, -1: 3, -0.25: -0.75, -0.5: 0},
+        ),
+    ],
+)
+def test_cheb_call_1D(cheb_function, degree, answer_key):
+    """Test chebyshev polynomial call with a 1D array"""
+    f = cheb_function(degree)
+    xs = [0.5, 1, -1, -0.25, -0.5]
```
mphoward: This could be changed to:
```suggestion
    xs = list(answer_key.keys())
```
and this would also work with trig, right?

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+            smolyay.basis.ChebyshevSecondKind,
+            2,
+            {0.5: 0, 1: 3, -1: 3, -0.25: -0.75, -0.5: 0},
+        ),
+    ],
+)
+def test_cheb_call_2D(cheb_function, degree, answer_key):
+    """Test chebyshev polynomial call with a 2D array"""
+    f = cheb_function(degree)
+    xs = [0.5, 1, -1, -0.25, -0.5, -1, 1, -0.25]
```
mphoward: Same here.

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+    assert f(0.5) == pytest.approx(answer_key[0.5])
+    assert f(1) == pytest.approx(answer_key[1])
+    assert f(-1) == pytest.approx(answer_key[-1])
+    assert f(-0.25) == pytest.approx(answer_key[-0.25])
+    assert f(-0.5) == pytest.approx(answer_key[-0.5])
+
+
+@pytest.mark.parametrize(
+    "cheb_function,degree,answer_key",
+    [
```
mphoward: Rather than repeat this list in multiple tests, I think you can save it in a variable once and then reuse it in all the parametrizes.

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+            smolyay.basis.ChebyshevSecondKind,
+            2,
+            {0.5: 0, 1: 3, -1: 3, -0.25: -0.75, -0.5: 0},
+        ),
+    ],
+)
+def test_cheb_call_3D(cheb_function, degree, answer_key):
+    """Test chebyshev polynomial call with a 3D array"""
+    f = cheb_function(degree)
+    xs = [
```
mphoward: Grab the keys, then you can use numpy.tile to repeat it if you want

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
     with pytest.raises(ValueError):
         f(-2)
     with pytest.raises(ValueError):
-        f([0.5,0.7,3,0.8])
+        f([0.5, 0.7, 3, 0.8])
+    with pytest.raises(ValueError):
+        f([[0.5, 0.7, 3, 0.8], [0.5, 0.7, 0.8, -0.5]])
+
 
+@pytest.mark.parametrize(
```
mphoward, line 317: Same ideas down here, you can make another list of derivatives and reuse it.

tests/basis_test.py
```diff
@@ -1,330 +1,1167 @@
...
+    f2 = smolyay.basis.Trigonometric(2)
+    assert f2.frequency == 2
+    assert isinstance(f2.frequency, int)
+    assert numpy.array_equal(f2.domain, [0, 2 * numpy.pi])
+    f2.frequency = float(3)
+    assert f2.frequency == 3
+    assert isinstance(f2.frequency, int)
+
+
+@pytest.mark.parametrize(
```
mphoward: I think all these tests can be incorporated into the ones above by adding `Trigonometric(0)`, etc. into the function test list above.

mphoward (changes requested):
Thanks Michaela! I like what you have done with the parametrization, but I would like to reduce the amount of code duplication. See my comments and let me know if you have questions. I think this will cut down on the size of this file quite a bit and make it easier to maintain / add new basis functions in future.

tests/basis_test.py
mzbush: I've created a global parametrized set for the calls and the derivatives that uses all the basis functions. The total number of lines has been extremely reduced. I've also done a similar thing with the tests for checking that the call and derivative raise an error for inputs outside the domain. The test file is now organized like this:

1. Parameters used by multiple tests
2. Initialization and class property tests
3. Error raising for outside domain tests
4. Call correctness tests
5. Derivative correctness tests
6. Correctness tests for points that are special to a basis function
7. Basis set initialization test

tests/basis_test.py
```diff
@@ -1,330 +1,438 @@
...
-    """Test call raises error if input is outside domain [-1,1]"""
-    f = ChebyshevFirstKind(4)
-    with pytest.raises(ValueError):
-        f.derivative(2)
+import smolyay
+
+
+basis_call_answer_key = pytest.mark.parametrize(
+    "basis_fun,answer_key",
+    [
```
mphoward: For readability, I would suggest you just make this a list, and then use the `parametrize` decorator on the tests itself. That way it's clear what's going on, and also those tests get to pick what they want to call the argument names.
```suggestion
basis_call_answer_key =
    [
```
Please do the same for the others like this.

tests/basis_test.py
```diff
@@ -1,330 +1,438 @@
...
+            ]
+        )
+
+
+# Test call function correctness
+@basis_call_answer_key
+def test_call(basis_fun, answer_key):
+    """Test basis function call"""
+    print(basis_fun)
+    print(answer_key)
```
mphoward: Remove print statements from tests. If there is an issue where the parametrized tests are not showing up with nice names, there is a way to get nicer names with pytest that we can look at!

mphoward (approved):
Thanks Michaela! This looks great, especially with the changes you've added to reduce duplication. I have only two minor comments, and then this can be merged if @cakieslich and @mzf0069 also approve.

tests/basis_test.py
mzbush: I've added ids to these tests and the benchmark function tests.

mzf0069 (approved):

cakieslich (approved):

PR #21, opened 2024-04-17
Unidimensional sample points

mzbush:
A series of classes for handling unidimensional points sets.


smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    The natural domain of the points specified by :param:``domain`` is
+    the domain in which all the possible points in the set lies within.
+    ``points`` is the list of points that the UnidimensionalPointSet
+    is expected to store.
+    ``domain`` is the domain of the points generated by the point set.
+    """
+
+    def __init__(self):
+        self._points = None
+        self._valid_cache = True
```
mphoward: A cache should begin invalid so that it will be filled the first time it's needed.
```suggestion
        self._valid_cache = False
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    in a tensor product grid or sparse grid. These unidimensional points are
+    associated with certain polynomial families, and the combination of
+    unique points and polynomial families are the basis for different types
+    of quadrature rules that are used in numerical integration and in
+    approximation.
+    The natural domain of the points specified by :param:``domain`` is
+    the domain in which all the possible points in the set lies within.
+    ``points`` is the list of points that the UnidimensionalPointSet
+    is expected to store.
+    ``domain`` is the domain of the points generated by the point set.
```
mphoward: I think most of this documentation is not current, so I would remove it for now.

Just FYI: attributes and properties are usually also documented where they are defined, rather than at the high level of the class, so that is why I'm suggesting to remove those here.
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    """
+
+    def __init__(self):
+        self._points = None
+        self._valid_cache = True
+
+    @property
+    @abc.abstractmethod
+    def domain(self):
+        """list: Domain the sample points come from."""
```
mphoward: Let's make this a numpy array like we did for the basis functions
```suggestion
        """numpy.ndarray: Domain of the `points`."""
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        self._valid_cache = False
+        return self._points
+
+    @abc.abstractmethod
+    def _create(self):
+        r"""Generating the points
+
+        An abstract method for generating the points stored by the point set
+        and other parameters.
+        """
```
```suggestion
        """Create the points in the set."""
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+
+    @property
+    @abc.abstractmethod
+    def domain(self):
+        """list: Domain the sample points come from."""
+        pass
+
+    @property
+    def points(self):
+        """list: Points stored by the set."""
```
```suggestion
        """numpy.ndarray: Points in the set."""
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        r"""Generating the points
+
+        An abstract method for generating the points stored by the point set
+        and other parameters.
+        """
+        pass
+
+
+class NestedUnidimensionalPointSet(UnidimensionalPointSet):
+    """Set of unidimensional points with levels
```
```suggestion
    """Set of unidimensional points assigned to levels.
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    of quadrature rules that are used in numerical integration and in
+    approximation.
+
+    The number of unique points is controlled by `num_points`, which describes
+    the amount of points added by each level.
+    `points` describes the unique points added as the number of levels increase,
+    and the order of elements is reflective of the level each element first
+    appears in. Some point set are nested, meaning that the points in a previous
+    level reappear in following levels. However, `points` does not attempt to
+    capture this behavior and should be free of duplicates.
```
mphoward: I would also remove this for now since I think it is a little out of date / documents attributes. We will want to revisit all documentation in a future PR anyway.
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+
+    @property
+    def max_level(self):
+        """int: maximum level to compute points for."""
+        return self._max_level
+
+    @max_level.setter
+    def max_level(self, value):
+        self._max_level = value
+        self._valid_cache = True
```
mphoward: When using a cache, usually you guard against invalidating the cache if the value hasn't actually changed. This is the pattern, also including a typecast / copy on `value`.
```suggestion
        max_level = int(value)
        if max_level != self._max_level:
            self._max_level = max_level
            self._valid_cache = False
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    def domain(self):
+        """list: Domain the sample points come from."""
+        pass
+
+    @property
+    def points(self):
+        """list: Points stored by the set."""
+        if self._valid_cache:
+            self._create()
+        self._valid_cache = False
```
mphoward: The pattern should be like this (if cache is invalid, fill it). Please propagate forward.
```suggestion
        if not self._valid_cache:
            self._create()
            self._valid_cache = True
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        return self._max_level
+
+    @max_level.setter
+    def max_level(self, value):
+        self._max_level = value
+        self._valid_cache = True
+
+    @property
+    def num_points(self):
+        """list: number of points per level."""
```
mphoward: I think all of these should probably be NumPy arrays.
```suggestion
        """numpy.ndarray: number of points per level."""
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        """int: maximum level to compute points for."""
+        return self._max_level
+
+    @max_level.setter
+    def max_level(self, value):
+        self._max_level = value
+        self._valid_cache = True
+
+    @property
+    def num_points(self):
```
mphoward: I think `num_points` might be confusing because there is also an attribute called `points`. What about `num_points_in_level` or `size_of_level`?

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        return numpy.array([-1, 1])
+
+    @property
+    def degree(self):
+        """int: degree of polynomial to create points for."""
+        return self._degree
+
+    @degree.setter
+    def degree(self, value):
+        self._degree = value
```
mphoward: Use same pattern here to guard the cache.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    of quadrature rules that are used in numerical integration and in
+    approximation.
+    The natural domain of the points specified by :param:``domain`` is
+    the domain in which all the possible points in the set lies within.
+    ``points`` is the list of points that the UnidimensionalPointSet
+    is expected to store.
+    ``domain`` is the domain of the points generated by the point set.
+    """
+
+    def __init__(self):
```
mphoward: I thought that these point sets would take a domain, and deriving classes would scale the points as needed? You will need to add an argument here for the domain, coerce it to a numpy array, and add a readonly property.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    level reappear in following levels. However, `points` does not attempt to
+    capture this behavior and should be free of duplicates.
+
+    Parameters
+    ----------
+    max_level ; int
+        the maximum level the points are used for
+    """
+
+    def __init__(self, max_level):
```
mphoward: Add the domain argument here before `max_level`, and pass `domain` through to the parent constructor.
```suggestion
    def __init__(self, max_level):
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        super().__init__()
+        self.max_level = max_level
+        self._num_points = None
+        self._start_level = None
+        self._end_level = None
```
mphoward: You will want to reorder like this so that the guard below works.
```suggestion
        self._max_level = None
        self._num_points = None
        self._start_level = None
        self._end_level = None

        self.max_level = max_level
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+
+    For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
+
+    Parameters
+    ----------
+    degree : int
+        degree of the Chebyshev polynomial of the first kind to get extrema from
+    """
+
+    def __init__(self, degree):
```
mphoward: Add domain to constructor. You can pick what order makes sense (domain first or not), then be consistent in all children

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        self._valid_cache = True
+
+    def _create(self):
+        r"""Generating the points
+
+        Generating the extrema of a Cheybshev polynomial of the first kind at
+        a given degree.
+        """
+        if self.degree > 0:
+            points = -numpy.cos(
```
mphoward, line 183: Scale these by the domain. You can add a helper function somewhere for doing that transformation since it will be needed in most children.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+    polynomial with degree 2^{k - 1} will be added followed by the extrema of the
+    polynomial with degree 2^{k} , then 2^{k + 1} , then 2^{k + 2} and so on.
+
+    Parameters
+    ----------
+    max_level ; int
+        the maximum level the points are used for
+    """
+
+    def __init__(self, max_level):
```
mphoward: Domain needed (please propagate forward)

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+            points = -numpy.cos(
+                numpy.pi * numpy.linspace(0, self.degree, self.degree + 1) / self.degree
+            )
+        else:
+            points = numpy.zeros(1)
+        self._points = points
+
+
+class NestedClenshawCurtisPointSet(NestedUnidimensionalPointSet):
+    r"""Generate nested Clenshaw Curtis points in accordance with a growth rule
```
mphoward: Update this documentation to remove mentions of other growth rules, and just define how this one works. You can also refer to the `ClenshawCurtisPointSet` for details of how the extrema are generated, and just focus on what the levels mean. That will probably be more helpful for a user.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+
+        Generating nested extrema of chebyshev polynomials of the first kind.
+        """
+        # create properties for levels
+        rule = lambda x: 1 if x == 0 else 2**x + 1
+        self._num_points = [rule(0)] + [
+            rule(i) - rule(i - 1) for i in range(1, self.max_level + 1)
+        ]
+        self._start_level = numpy.cumsum([0] + list(self._num_points))[:-1]
+        self._end_level = numpy.cumsum(list(self._num_points))
```
mphoward: Will this work instead? It seems easier to read to me
```suggestion
        self._num_points = numpy.ones(self.max_level + 1, dtype=int)
        self._num_points[1:] = [2**i - 2**(i-1) for i in range(1, self.max_level + 1)]

        self._start_level = numpy.zeros_like(self._num_points)
        self._start_level[1:] = numpy.cumsum(self._num_points)

        self._end_level = self._start_level + self._num_points
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        # points
+        points = [0]
+        degree = 0
+        num_levels = numpy.sum(numpy.array(self._num_points) != 0)
```
mphoward: This is known a priori, right? Also, this variable should also be used to size the arrays above, so probably you want to define it at the top of this function.
```suggestion
        num_levels = self.max_level + 1
```

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        for i in range(1, num_levels):
+            degree = 2**i
+            if i == 1:
+                indexes = numpy.linspace(0, degree, 2, dtype=int)
+            else:
+                indexes = numpy.linspace(1, degree - 1, degree - 1, dtype=int)
```
mphoward: I'm confused by the use of `linspace` here. Should this be `arange`?

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
```
mphoward: A comment on what this code is doing would be helpful, and possibly putting the selection logic into a variable to make it more obvious what's going on.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+            new_points = list(-numpy.cos(numpy.pi * indexes / degree))
+            points.extend(new_points)
```
mphoward: Size the `points` memory before here since you know the total number of points (can do `numpy.sum` on `_num_points`), then write into that array using the start and end indexes. That way you don't need to cast to list.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+        # create properties for levels
+        rule = lambda x: 1 if x == 0 else 2**x + 1
+        self._num_points = [rule(0)] + [
+            rule(i) - rule(i - 1) for i in range(1, self.max_level + 1)
+        ]
+        self._start_level = numpy.cumsum([0] + list(self._num_points))[:-1]
+        self._end_level = numpy.cumsum(list(self._num_points))
+        # points
+        points = [0]
+        degree = 0
```
mphoward: See comment below about sizing points.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
+            new_points = list(-numpy.cos(numpy.pi * indexes / degree))
+            points.extend(new_points)
+        self._points = points
+
+
+class SlowNestedClenshawCurtisPointSet(NestedClenshawCurtisPointSet):
+    """Set for Clenshaw Curtis slow exponential growth"""
+
+    def _create(self):
```
mphoward, line 333: After we converge on the nested CC points above, apply the same changes down here. You can hold off on doing this now for efficiency.

smolyay/samples.py
```diff
@@ -0,0 +1,362 @@
...
+                indexes = numpy.linspace(0, degree, 2, dtype=int)
+            else:
+                indexes = numpy.linspace(1, degree - 1, degree - 1, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
+            new_points = list(-numpy.cos(numpy.pi * indexes / degree))
+            points.extend(new_points)
+        self._points = points
+
+
+class TrigonometricPointSet(UnidimensionalPointSet):
```
mphoward, line 270: Same comments about updating these, but wait until we're good on the CC points.

mphoward (changes requested):
Thanks Michaela! This is generally in good shape for a first pass. I didn't review the tests yet, but I had some comments on the abstract classes and Clenshaw-Curtis points that we can work on, then apply the same changes to the other point sets.

smolyay/samples.py
mzbush: No, when calculating the number of points per level, the first level is one less than it should be because 2**(i-1) is 1 when i = 1.

mzbush: I was able to make adjustments so it would work.

smolyay/samples.py
mzbush: I like `num_points_per_level`

smolyay/samples.py
mzbush: I was under the impression that it was the MultidimensionalPointSet that would be handling the domain and scaling.

mphoward: No, that won’t work with the interface we discussed with Dr. K at our last meeting to build a multidimensional set from a list of unidimensional sets. The unidimensional sets need to already have their domains set for that to work.

mzbush: Okay, I went ahead and added domain as a parameter.

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+
+class UnidimensionalPointSet(abc.ABC):
+    """Set of unidimensional points
+
+    A set of unique unidimensional points within a domain.
+    """
+
+    def __init__(self, domain):
+        self._points = None
+        self._domain = numpy.array(domain)
```
mphoward: You should make sure to cast the type, in case they are given as ints or something.
```suggestion
        self._domain = numpy.array(domain, dtype=float)
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+    def domain(self):
+        """numpy.ndarray: Domain of the `points`."""
+        return self._domain
+
+    @property
+    def points(self):
+        """numpy.ndarray: Points in the set."""
+        if not self._valid_cache:
+            self._create()
+        self._valid_cache = True
```
mphoward: Put this inside the if, it only needs to run if the cache was filled.
```suggestion
            self._valid_cache = True
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        return self._points
+
+    @abc.abstractmethod
+    def _create(self):
+        """Create the points in the set."""
+        pass
+
+    @staticmethod
+    def scale_to_domain(x, old, new):
```
mphoward: I think this should be a normal method so that it can access the domain. I think it should also be private, since I don't think we want it to be part of the public API in case we change our mind.
```suggestion
    def _scale_to_domain(self, points, old_domain):
```
Last, please remove print statements used for debugging.

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+    def __init__(self, domain, degree):
+        super().__init__(domain)
+        self._degree = None
+
+        self.degree = degree
+
+    @property
+    def domain(self):
+        """numpy.ndarray: Domain the sample points come from."""
+        return numpy.array([-1, 1])
```
mphoward: Remove this, you're accidentally overriding the parent class property.
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+
+    A set of unique unidimensional points within a domain.
+    """
+
+    def __init__(self, domain):
+        self._points = None
+        self._domain = numpy.array(domain)
+        self._valid_cache = False
+
+    @property
```
mphoward, line 24: Looking at the code, I think it would be OK to add a setter for the domain that invalidates the point cache. We can worry upstream (in the multivariable point sets or the surrogates) about using these in an immutable way.

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        rule = lambda x: 1 if x == 0 else 2**x + 1
+        self._num_points_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_points_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+
+        self._start_level = numpy.zeros_like(self._num_points_per_level)
+        self._start_level[1:] = numpy.cumsum(self._num_points_per_level[:-1])
+
+        self._end_level = self._start_level + self._num_points_per_level
```
mphoward: I might have had this backwards before, but maybe this will work and be easier?
```suggestion
        self._end_level = numpy.cumsum(self._num_points_per_level)
        self._start_level = self._end_level - self._num_points_per_level
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        # points
+        points = numpy.zeros(numpy.sum(self._num_points_per_level))
```
mphoward: The total number of points should be the last value in end level.
```suggestion
        num_points = self._end_level[-1]
        points = numpy.zeros(num_points, dtype=float)
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        degree = 0
```
mphoward: This variable is unused if the loop is not entered, right?
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        num_levels = self.max_level + 1
+        for i in range(1, num_levels):
```
```suggestion
        for i in range(1, self.max_level + 1):
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+            # find indexes of extrema not already found. Fraction index/degree
+            # cannot be further simplified
+            if i == 1:
+                indexes = numpy.arange(0, degree + 1, 2, dtype=int)
+            else:
+                indexes = indexes = numpy.arange(1, degree, 2, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
+            # generate extrema
+            new_points = list(-numpy.cos(numpy.pi * indexes / degree))
+            points[self._start_level[i] : self._end_level[i]] = new_points
```
mphoward: No need to go through a list
```suggestion
            points[self._start_level[i] : self._end_level[i]] = -numpy.cos(numpy.pi * indexes / degree)
```

smolyay/samples.py
```diff
@@ -0,0 +1,409 @@
...
+        num_levels = self.max_level + 1
+        for i in range(1, num_levels):
+            degree = 2**i
+            # find indexes of extrema not already found. Fraction index/degree
+            # cannot be further simplified
+            if i == 1:
+                indexes = numpy.arange(0, degree + 1, 2, dtype=int)
+            else:
+                indexes = indexes = numpy.arange(1, degree, 2, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
```
```suggestion
                indexes = numpy.arange(1, degree, 2, dtype=int)
                divisible_indexes = numpy.gcd(indexes, degree) > 1
                indexes = indexes[~divisible_indexes]
```

mphoward (changes requested):
Thanks! I think you hit most of the things requested. Here are some more comments on the code we've gone over so far. Let me know if you have questions and then if this seems good to go after you've made those changes, you can propagate the changes to the other two types of point sets.

smolyay/samples.py
mzbush: Would we really want to recalculate all the all after a domain change? We could just scale the existing points to the new domain.

mphoward: This implementation is the quickest to code (although maybe inefficient), so I would do it this way and then you can always optimize it later if you find it’s an issue with profiling.

smolyay/samples.py
mzbush: maybe `num_per_level`?

smolyay/samples.py
mzbush: okay

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+
+    A set of unique unidimensional points within a domain.
+    """
+
+    def __init__(self, domain):
+        self._points = None
+        self._domain = None
+
+        self.domain = numpy.array(domain, dtype=float)
+        self._valid_cache = False
```
mphoward: I would group valid cache with the other internal variables, and then the domain setter should take care of the casting and copying now.
```suggestion
        self._valid_cache = False

        self.domain = domain
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @property
+    def domain(self):
+        """numpy.ndarray: Domain of the `points`."""
+        return self._domain
+
+    @domain.setter
+    def domain(self, value):
+        domain = numpy.array(value, dtype=float)
+        if domain.ndim != 1 or len(domain) != 2:
```
mphoward: I think you can use a tuple comparison here:
```suggestion
        if domain.shape != (2,):
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @max_level.setter
+    def max_level(self, value):
+        max_level = int(value)
+        if max_level != self._max_level:
+            self._max_level = max_level
+            self._valid_cache = False
+
+    @property
+    def num_points_per_level(self):
```
mphoward: I think you had suggested this in a comment before? I liked this idea to shorten the name
```suggestion
    def num_per_level(self):
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        r"""Create the points in the set.
+
+        Generating the extrema of a Cheybshev polynomial of the first kind at
+        a given degree.
+        """
+        if self.degree > 0:
+            points = -numpy.cos(
+                numpy.pi * numpy.linspace(0, self.degree, self.degree + 1) / self.degree
+            )
+        else:
```
mphoward, line 186: I've tripped myself reading this code a few times because of the `>`. A comment here like this would help clarify:
```suggestion
        else:
            # special case where degree == 0
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        return points
+
+
+class NestedUnidimensionalPointSet(UnidimensionalPointSet):
+    """Set of unidimensional points assigned to levels.
+
+
+    Parameters
+    ----------
+    max_level ; int
```
mphoward: Be sure to document all arguments of a constructor, even if they get forwarded to a parent class.

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+    and k < L.
+
+    Parameters
+    ----------
+    max_level ; int
+        the maximum level the points are used for
+    """
+
+    def __init__(self, domain, max_level):
+        super().__init__(domain, max_level)
```
mphoward: You can omit a constructor if all it does is call the parent.
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        # create properties for levels
+        rule = lambda x: 1 if x == 0 else 2**x + 1
+        self._num_points_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_points_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+
+        self._end_level = numpy.cumsum(self._num_points_per_level)
+        self._start_level = self._end_level - self._num_points_per_level
```
mphoward: Suggest adding a comment and using white space to organize here
```suggestion
        # create properties for levels, level 0 is a special case with 1 point
        rule = lambda x: 1 if x == 0 else 2**x + 1
        self._num_points_per_level = numpy.ones(self.max_level + 1, dtype=int)
        self._num_points_per_level[1:] = [
            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
        ]
        self._end_level = numpy.cumsum(self._num_points_per_level)
        self._start_level = self._end_level - self._num_points_per_level

```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        # points
```
```suggestion
        # points, level 0 is a special case only 0 as a point
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        num_points = self._end_level[-1]
+        points = numpy.zeros(num_points, dtype=float)
+        for i in range(1, self.max_level + 1):
+            degree = 2**i
+            # find indexes of extrema not already found. Fraction index/degree
+            # cannot be further simplified
```
```suggestion
            # find indexes of extrema not already found. Fraction index/degree
            # cannot be further simplified
            degree = 2**i
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+            if i == 1:
+                indexes = numpy.arange(0, degree + 1, 2, dtype=int)
+            else:
+                indexes = numpy.arange(1, degree, 2, dtype=int)
+                divisible_indexes = numpy.gcd(indexes, degree) > 1
+                indexes = indexes[~divisible_indexes]
+            # calculate new points
```
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+    points remains linear with respect to L.
+
+    """
+
+    def _create(self):
+        r"""Create the points in the set.
+
+        Generating nested extrema of chebyshev polynomials of the first kind.
+        """
+        # create properties for levels
```
mphoward: Make this method look like the `ClenshawCurtisPointSet` with the suggested changes. It would be good to also add some explanation for where the formula for the rule comes from.

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+
+        self._end_level = numpy.cumsum(self._num_points_per_level)
+        self._start_level = self._end_level - self._num_points_per_level
+        # points
+        points = numpy.zeros(numpy.sum(self._num_points_per_level))
+        degree = 0
+        num_levels = numpy.sum(numpy.array(self._num_points_per_level) != 0)
+        nonempty_index = lambda x: 0 if x == 0 else int(2 ** (x - 2)) + 1
```
mphoward: Is it necessary to do this? I would think you could just loop over all the levels and use `continue` if the level is empty. You can increment a counter of non-empty levels if needed for doing the integer math.

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        r"""Create the points in the set.
+
+        Generating trigonometic points at a given frequency.
+        """
+        if self.frequency > 0:
+            idx = numpy.arange(1, self.frequency + 1, 1, dtype=int)
+            points = (idx - 1) * 2 * numpy.pi / self.frequency
+        else:
+            points = numpy.zeros(1)
+        self._points = points
```
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,415 @@
...
+        self._points = self._scale_to_domain(points, [0, 2 * numpy.pi])
+
+
+class NestedTrigonometricPointSet(NestedUnidimensionalPointSet):
```
mphoward, line 439: Please also apply changes to this class now.

tests/samples_test.py
```diff
@@ -0,0 +1,405 @@
...
+                5 / 9,
+                7 / 9,
+                8 / 9,
+            ]
+        ),
+    ),
+]
+
+sample_points_ids = [
+    "Non-nested Cheb [0]",
```
mphoward: I would name these to match the point sets we have and don't include "non-nested", like
```suggestion
    "ClenshawCurtis [0]",
```

tests/samples_test.py
```diff
@@ -0,0 +1,405 @@
...
+    "Nested Trig [2]",
+]
+
+
+def test_clenshaw_initial():
+    """test default properties"""
+    f = smolyay.samples.ClenshawCurtisPointSet([-2, 1], 7)
+    assert numpy.array_equal(f.domain, [-2, 1])
+    assert f.degree == 7
+    assert f._valid_cache == False
```
mphoward: I don't recommend including private components of the API in testing unless absolutely necessary. Better would be to have a test later where you get the points with some initial parameter value, change that value, and then check that the points update correctly
```suggestion
```

mphoward (changes requested):
Nice work Michaela! These tests look really clean, and I had only some small comments remaining to address (mostly to propagate the changes forward to the classes that we held off on). Once you implement those, I am good to approve this PR.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+            self._create()
+            self._valid_cache = True
+        return self._end_level
+
+
+class ClenshawCurtisPointSet(UnidimensionalPointSet):
+    r"""Set of unidimensional points for Clenshaw Curtis sampling
+
+    The :attr:`points` for this interpolation scheme are the extrema of the
+    Chebyshev polynomials of the first kind on the domain :math:`[-1, 1]`:
```
mphoward, line 136: Clarify here that these points are then scaled to the specified domain.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    r"""Generate nested Clenshaw Curtis points
+
+    The :attr:`points` for this interpolation scheme are the extrema of the
+    Chebyshev polynomials of the first kind on the domain :math:`[-1, 1]`:
+
+    .. math::
+
+        x_i^* = -\cos(\pi i/n), i = 0,...,n
+
+    For the special case :math:`n = 0`, there is only one point :math:`x_0^* = 0`.
```
mphoward: This documentation is a bit redundant of ClenshawCurtisPointSet, so I would probably refer the reader there and focus on explaining how the levels / nesting are defined & related to those points.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+
+    .. math::
+        num_per_level(L) = o(L) - o(L - 1)
+
+    Parameters
+    ----------
+    domain: list
+        Domain of the sample points.
+
+    max_level ; int
```
```suggestion
    max_level: int
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+            points[self._start_level[i] : self._end_level[i]] = -numpy.cos(
+                numpy.pi * indexes / degree
+            )
+        self._points = self._scale_to_domain(points, [-1, 1])
+
+
+class SlowNestedClenshawCurtisPointSet(NestedClenshawCurtisPointSet):
+    r"""Set for Clenshaw Curtis slow exponential growth.
+
+    The :attr:`points` for this interpolation scheme are the extrema of the
```
mphoward: Same comments here about making the docs focus on what's special about this class compared to ClenshawCurtisPointSet and NestedClenshawCurtisPointSet.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+                indexes = numpy.arange(1, degree, 2, dtype=int)
+                divisible_indexes = numpy.gcd(indexes, degree) > 1
+                indexes = indexes[~divisible_indexes]
+            points[self._start_level[i] : self._end_level[i]] = -numpy.cos(
+                numpy.pi * indexes / degree
+            )
+        self._points = self._scale_to_domain(points, [-1, 1])
+
+
+class SlowNestedClenshawCurtisPointSet(NestedClenshawCurtisPointSet):
```
mphoward: Should this inherit from NestedClenshawCurtisPointSet or NestedUnidimensionalPointSet? I would say the second one, since a SlowNestedClenshawCurtisPointSet is not really a NestedClenshawCurtisPointSet in the way we have defined them, and it doesn't actually use anything from that class either.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    
+    .. math::
+        num_per_level(L) = o(L) - o(L - 1)    
+
+    Parameters
+    ----------
+    domain: list
+        Domain of the sample points.
+
+    max_level ; int
```
```suggestion
    max_level: int
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+
+class TrigonometricPointSet(UnidimensionalPointSet):
+    r"""Set of unidimensional points for Trigonometric sampling
+
+    The :attr:`points` for this interpolation scheme are nested
+    trigonometric points
+
+    .. math::
+
+        x^l_j = \frac{j-1}{m(l)},  1 \leq j \leq m(l), l \geq 0
```
mphoward: This documentation needs updating, it is referring to the nested version. This should probably be saying something about being the roots / extrema of a certain trig function & frequency, wherever it is these come from.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    .. math::
+        num_per_level(L) = o(L) - o(L - 1)
+
+
+    Parameters
+    ----------
+    domain: list
+        Domain of the sample points.
+
+    max_level ; int
```
```suggestion
    max_level: int
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        """
+        # create properties for levels, level 0 is a special case with 1 point
+        rule = lambda x: 3**x
+        self._num_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+
+        self._end_level = numpy.cumsum(self._num_per_level)
+        self._start_level = self._end_level - self._num_per_level
```
```suggestion
        self._end_level = numpy.cumsum(self._num_per_level)
        self._start_level = self._end_level - self._num_per_level

```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        # points, level 0 is a special case only 0 as a point
+        points = numpy.zeros(numpy.sum(self._num_per_level))
+        degree = 0
```
```suggestion
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        self._num_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+        self._end_level = numpy.cumsum(self._num_per_level)
+        self._start_level = self._end_level - self._num_per_level
+
+        # points, level 0 is a special case only 0 as a point
+        points = numpy.zeros(numpy.sum(self._num_per_level))
+        degree = 0
```
```suggestion
```

tests/samples_test.py
```diff
@@ -0,0 +1,420 @@
+import pytest
+
+import numpy
```
mphoward: I think these should be ordered like this according to PEP 8
```suggestion
import numpy
import pytest
```

tests/samples_test.py
```diff
@@ -0,0 +1,420 @@
...
+
+import smolyay.samples
```
mphoward: Just import smolyay to make sure the internal package is working as intended
```suggestion
import smolyay
```

tests/samples_test.py
```diff
@@ -0,0 +1,420 @@
...
+    new_domain = samples.domain * 5 + 10
+    samples.domain = new_domain
+    assert numpy.array_equal(samples.domain, new_domain)
+    assert numpy.allclose(samples.points, new_points, atol=1e-10)
+
+
+@pytest.mark.parametrize("samples,points", sample_points_answers, ids=sample_points_ids)
+def test_len(samples, points):
+    """test the len method of UnidimensionalPointSet objects"""
+    assert len(samples) == len(points)
```
mphoward: These are all related, so I would combine these three tests to:
1. Check the length of samples match.
2. Check the points match before the domain is rescaled.
3. Check the points match after the domain is rescaled.

tests/samples_test.py
```diff
@@ -0,0 +1,420 @@
...
+    with pytest.raises(TypeError):
+        samples([-10, 10, 20], 7)
+    with pytest.raises(TypeError):
+        samples([[-10, 10], [-10, 10]], 7)
+
+
+@pytest.mark.parametrize("samples,points", sample_points_answers, ids=sample_points_ids)
+def test_generate_points(samples, points):
+    """test the points of initialized UnidimensionalPointSet"""
+    assert numpy.allclose(samples.points, points, atol=1e-10)
```
mphoward, line 306: Is atol actually needed here?

tests/samples_test.py
```diff
@@ -0,0 +1,420 @@
...
+        ),
+        (smolyay.samples.NestedTrigonometricPointSet([0, 2 * numpy.pi], 0), [1]),
+        (smolyay.samples.NestedTrigonometricPointSet([0, 2 * numpy.pi], 1), [1, 3]),
+        (smolyay.samples.NestedTrigonometricPointSet([0, 2 * numpy.pi], 2), [1, 3, 9]),
+    ],
+    ids=nested_sample_ids,
+)
+def test_nested_end_level(nested_samples, end_level):
+    """test number of points per level"""
+    assert numpy.array_equal(nested_samples.end_level, end_level)
```
mphoward: You can combine these tests, it is fine to run multiple related asserts in a test

mphoward (approved):
Thanks Michaela! I have a couple final comments, and then I'm good with this.

smolyay/samples.py
mzbush: It originally inherited from the NestedClenshawCurtisPointSet under the assumption that calculating certain things like points would be separate methods that the slow version could inherit, It should inherit from the NestedUnidimensionalPointSet since that's not the case anymore.

tests/samples_test.py
mzbush: The default atol was too lenient.

tests/samples_test.py
mzbush: I would ideally like to combine the `start_level`, `end_level`, and the `get_individual_levels` tests since the `get_individual_levels` test tests whether the values in the start_level and end_level get you the expected points for each level (basically testing if the `start_level` and `end_level` variables fulfill their purpose). However, the list of parameters in the `get_individual_levels` test isn't compatible with the `start_level` and `end_level` tests, so I'm not sure if there's a good way to do this.

mzbush: The parameters for `get_individual_levels` could be made the same as `start_level`, `end_level`, and `num_per_level`, but that would mean always checking the last level of the object. Alternatively, I could have the `num_per_level`, `start_level`, and `end_level`, tested for correctness, save the current `max_level` to a reference variable `check_level`, change the `max_level` to some constant, and then check the points at the level in `check_level`.

This isn't a test that check if `max_level` is set correctly, and in fact there is no such test for checking that changing `max_level` causes the points to update.

tests/samples_test.py
```diff
@@ -0,0 +1,448 @@
...
+    ids=nested_sample_ids,
+)
+def test_nested_levels(nested_samples, num_per_level, start_level, end_level):
+    """test number of points per level and start and end indexes"""
+    assert numpy.array_equal(nested_samples.num_per_level, num_per_level)
+    assert numpy.array_equal(nested_samples.start_level, start_level)
+    assert numpy.array_equal(nested_samples.end_level, end_level)
+    check_level = nested_samples.max_level
+    nested_samples.max_level = 6
+    assert check_level != 6
```
mphoward: I'm not really sure what this is doing / how much it helps. If you want to check that changing `max_level` is working right, I would probably do a separate test (parametrized on just 1 instance of each type of nested set) and make sure that the number of points and the size of all the level-based variables changes correctly when you change the level.

tests/samples_test.py
```diff
@@ -0,0 +1,448 @@
...
+            ),
+        ),
+    ],
+    ids=nested_sample_ids,
+)
+def test_nested_get_individual_levels(nested_samples, level, points):
+    """test number of points per level"""
+    end_lev = nested_samples.end_level[level]
+    start_lev = nested_samples.start_level[level]
+    assert numpy.allclose(nested_samples.points[start_lev:end_lev], points)
```
mphoward: I didn't catch this before, but what is actually the purpose of this test? You already have a test that checks that the points are generated correctly for each type of point set, and you also have a test that makes sure the nested levels are defined correctly. Hence, I think all this is doing is making sure that Python's slice operator works, which we don't need to test. I would remove this.

tests/samples_test.py
mzbush: Added tests for setting `max_level`, `degree`, and `frequency` to the initialization tests. I also implemented the strategy for merging the `start_level`, `end_level`, and `num_per_level` test with the test that checks that the values in `start_level` and `end_level` do correspond to the points expected in the individual levels.

tests/samples_test.py
mzbush: That is because it was a partial implementation of a change that I thought I hadn't included in the commit because I was planning to be a full implementation in the morning.

tests/samples_test.py
mzbush: That is true that it's testing python's slice operator, but the main purpose is to check that what I think the `start_level` and `end_level` are supposed to be actually results in the desired behavior. It would very bad if what I or someone else assumed would get the right points in each level actually didn't, and the `start_level` and `end_level` calculations were fitted to match something that wasn't right. It essentially tests both the class and the answers to the previous test.

mphoward: > It essentially tests both the class and the answers to the previous test.

If this is the only purpose of this test, it should be removed because it duplicates the others.

mzbush: > If this is the only purpose of this test, it should be removed because it duplicates the others.

So we should either test that the `start_level` and `end_level` equal some predetermined lists or test that `samples.points[samples.start_level[i] : samples.end_level[i]]` nets you the expected points in level `i`, but not both?

If so, I would actually prefer to do the latter than the former.

mphoward: > So we should either test that the `start_level` and `end_level` equal some predetermined lists

You have already implemented this test, it's `test_nested_levels`. And `test_generate_points` checks that the points are created in the right order. So, there is no need to check slicing a list of points (that are known to be correct) between two indices (that are known to be correct).

mzbush: I've already implemented both checks, so I can remove either. I would prefer keeping the test where there's no risk of me or anyone else making an off-by-one error, and that's the test that checks that `start_level` and `end_level` get the right slicing behavior and not the test that checks if they equal certain lists.

mphoward: I disagree with this because it leads to significant code duplication and is a confusing / not helpful test for the reasons I outlined above. Please do as I asked.

mzbush: okay

smolyay/samples.py
mzbush: @mzf0069 Where do these come from?

mzf0069: https://arxiv.org/abs/1908.10672  

This is the reference for the trig function. I don't know what each point represents. The only note here is that the number of points at each level needs to be odd.

mzbush: > The only note here is that the number of points at each level needs to be odd.

Is that total number of points or unique points at each individual level? Because the total number of points is odd, the number of unique points is not.

@mphoward It's not much, but I did expand the docstrings.

smolyay/samples.py
```diff
@@ -0,0 +1,484 @@
...
+    @property
+    def end_level(self):
+        """numpy.ndarray: the ending index of each level."""
+        if not self._valid_cache:
+            self._create()
+            self._valid_cache = True
+        return self._end_level
+
+    def level(self, index):
+        """return all the points in a level"""
```
```suggestion
        """numpy.ndarray: Points in a level"""
```

smolyay/samples.py
```diff
@@ -0,0 +1,484 @@
...
+        rule = lambda x: 1 if x == 0 else int(2 ** (numpy.ceil(numpy.log2(x)) + 1) + 1)
+        self._num_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+        self._end_level = numpy.cumsum(self._num_per_level)
+        self._start_level = self._end_level - self._num_per_level
+
+        # points, level 0 is a special case only 0 as a point
+        points = numpy.zeros(numpy.sum(self._num_per_level))
```
```suggestion
        num_points = self._end_level[-1]
        points = numpy.zeros(num_points, dtype=float)
```

smolyay/samples.py
```diff
@@ -0,0 +1,484 @@
...
+        rule = lambda x: 3**x
+        self._num_per_level = numpy.ones(self.max_level + 1, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+        self._end_level = numpy.cumsum(self._num_per_level)
+        self._start_level = self._end_level - self._num_per_level
+
+        # points, level 0 is a special case only 0 as a point
+        points = numpy.zeros(numpy.sum(self._num_per_level))
```
```suggestion
        num_points = self._end_level[-1]
        points = numpy.zeros(num_points, dtype=float)
```

smolyay/samples.py
```diff
@@ -0,0 +1,484 @@
...
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.max_level + 1)
+        ]
+        self._end_level = numpy.cumsum(self._num_per_level)
+        self._start_level = self._end_level - self._num_per_level
+
+        # points, level 0 is a special case only 0 as a point
+        num_points = self._end_level[-1]
+        points = numpy.zeros(num_points, dtype=float)
+        for i in range(1, self.max_level + 1):
```
mphoward: One thing that I'm noticing looking at all of these is that we are defining `max_level` as the inclusive upper bound, even though we are starting from level 0. That means that there are `max_level + 1` levels. Is this the behavior we want? Our usual conventions of levels start from level 1, meaning that there are `max_level` levels.

It could potentially make more sense to have a `num_levels` instead, as this unambiguously then runs from `0, ..., num_levels-1`.

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+    "SlowNestedClenshawCurtis [3]",
+    "SlowNestedClenshawCurtis [4]",
+    "SlowNestedClenshawCurtis [5]",
+    "NestedTrigonometric [0]",
+    "NestedTrigonometric [1]",
+    "NestedTrigonometric [2]",
+]
+
+
+def test_clenshaw_initial():
```
mphoward:
```suggestion
def test_initialize_clenshaw():
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+    f = smolyay.samples.ClenshawCurtisPointSet([-2, 1], 7)
+    assert numpy.array_equal(f.domain, [-2, 1])
+    assert f.degree == 7
+    assert isinstance(f.degree, int)
+    f.degree = float(5)
+    assert f.degree == 5
+    assert isinstance(f.degree, int)
+
+
+def test_trig_initial():
```
mphoward:
```suggestion
def test_initialize_trig():
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+        smolyay.samples.SlowNestedClenshawCurtisPointSet,
+        smolyay.samples.NestedTrigonometricPointSet,
+    ],
+    ids=[
+        "NestedClenshawCurtis",
+        "SlowNestedClenshawCurtis",
+        "NestedTrigonometric",
+    ],
+)
+def test_nested_initial(nested_samples):
```
mphoward:
```suggestion
def test_initialize_nested(nested_samples):
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+    ],
+    ids=[
+        "ClenshawCurtis",
+        "Trigonometric",
+        "NestedClenshawCurtis",
+        "SlowNestedClenshawCurtis",
+        "NestedTrigonometric",
+    ],
+)
+def test_domain_error(samples):
```
mphoward: You should check that assigning the domain invalidly also works.

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+    assert isinstance(f.max_level, int)
+    f.max_level = float(5)
+    assert f.max_level == 5
+    assert isinstance(f.max_level, int)
+
+
+@pytest.mark.parametrize(
+    "samples",
+    [
+        smolyay.samples.ClenshawCurtisPointSet,
```
mphoward: I suggest you pass the arguments needed to make the set as an extra parameter like this, then unpack them as keyword args below. That way, you don't need to rely on all the sets having exactly the same constructor signature. I had to read this a couple times to figure out how it worked.
```suggestion
        (smolyay.samples.ClenshawCurtisPointSet, {"degree": 7})
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+        "Trigonometric",
+        "NestedClenshawCurtis",
+        "SlowNestedClenshawCurtis",
+        "NestedTrigonometric",
+    ],
+)
+def test_domain_error(samples):
+    """test error given invalid domain"""
+    with pytest.raises(TypeError):
+        samples([-10, 10, 20], 7)
```
mphoward: If you call the parameter `set_args`, you could do like this, for example:
```suggestion
        samples(domain=[-10, 10, 20], **set_args)
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+
+    # test for different domain and that points update after changing domain
+    new_points = numpy.array(points) * 5 + 10
+    new_domain = samples.domain * 5 + 10
+    samples.domain = new_domain
+    assert numpy.array_equal(samples.domain, new_domain)
+    assert numpy.allclose(samples.points, new_points, atol=1e-10)
+    assert numpy.allclose(samples[:], new_points, atol=1e-10)
+    for p, a in zip(samples, new_points):
+        assert p == pytest.approx(a)
```
mphoward: I would add an explicit single element check, and then your for-loop (testing that it works as an iterator) can be written more compactly by turning it into a list.
```suggestion
    assert samples[0] == pytest.approx(new_points[0])
    assert numpy.allclose(samples[:], new_points, atol=1e-10)
    assert numpy.allclose(list(samples), new_points, atol=1e-10)
```

tests/samples_test.py
```diff
@@ -0,0 +1,359 @@
...
+    ids=nested_sample_ids,
+)
+def test_nested_levels(nested_samples, num_per_level, start_level, end_level):
+    """test number of points per level, start level indexes, and end level indexes"""
+    assert numpy.array_equal(nested_samples.num_per_level, num_per_level)
+    assert numpy.array_equal(nested_samples.start_level, start_level)
+    assert numpy.array_equal(nested_samples.end_level, end_level)
+    for i in range(len(start_level)):
+        points = nested_samples[start_level[i] : end_level[i]]
+        assert numpy.allclose(nested_samples.level(i), points)
```
mphoward: You can use enumerate and zip here:
```suggestion
    for level, (start, end) in enumerate(zip(nested_samples.start_level, nested_samples.end_level)):
        assert numpy.allclose(nested_samples.level(level), nested_samples[start:end])
```

mphoward (changes requested):
Thanks Michaela! I like the new convenience methods you added. I had one open question regarding whether we should be using `max_level` or `num_levels` to define the nested sets, and otherwise I'm good with this after you apply the changes I've marked.

smolyay/samples.py
mzbush: `max_level` is equivalent to the level of exactness that we've been using in all the documentation and papers, so changing it could be more confusing. Surrogates with only level 0 have one sampling point, and I would much prefer to have 0 be the trivial case with a single point than for 1 to be the trivial case.

Doing it by `num_levels` would also mean there is no behavior for an input of 0, or if there is behavior, it would be to return an empty array.

mphoward: > `max_level` is equivalent to the level of exactness that we've been using in all the documentation and papers, so changing it could be more confusing.

I think we had discussed with Dr. K, and I was talking with Mohammadreza today, about how for the general case of mixed basis functions / anisotropic sampling there isn't as strong of a concept of level of exactness per dimension.

> Surrogates with only level 0 have one sampling point, and I would much prefer to have 0 be the trivial case with a single point than for 1 to be the trivial case.

Why?  FWIW, our paper has ks that are 1-indexed, so 0 is never allowed as a level. We have made them be 0-indexed in Python because that's the convention for this programming language, and so I find it pretty confusing to have the upper bound be inclusive. It's an anti-pattern for 0-indexed programming languages. We could still use `max_level` but make it exclusive, if you like that better. It's functionally the same as naming it `num_levels`, but I think `num_levels` is less ambiguous.

> Doing it by `num_levels` would also mean there is no behavior for an input of 0, or if there is behavior, it would be to return an empty array.

It probably doesn't make sense from a practical point of view to construct a set with `num_levels = 0`, but it would be an empty set if it were allowed.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    def domain(self):
+        """numpy.ndarray: Domain of the `points`."""
+        return self._domain
+
+    @domain.setter
+    def domain(self, value):
+        domain = numpy.sort(numpy.array(value, dtype=float))
+        if domain.shape != (2,):
+            raise TypeError("Domain must be array with two variables")
+        if domain[0] == domain[1]:
```
mphoward: I think it doesn't hurt to make this `>=` if we're checking anyway:
```suggestion
        if domain[0] >= domain[1]:
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        super().__init__(domain)
+        self._num_level = None
+        self._num_per_level = None
+        self._start_level = None
+        self._end_level = None
+
+        self.num_level = num_level
+
+    @property
+    def num_level(self):
```
mphoward: I would make this plural everywhere, since it reads a little strangely as "number of level"
```suggestion
    def num_levels(self):
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    def _create(self):
+        r"""Create the points in the set.
+
+        Generating nested extrema of chebyshev polynomials of the first kind.
+        """
+        # create properties for levels, level 0 is a special case with 1 point
+        rule = lambda x: 1 if x == 0 else 2**x + 1
+        self._num_per_level = numpy.ones(self.num_level, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.num_level)
```
```suggestion
            rule(i) - rule(i - 1) for i in range(1, self.num_level)
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+                continue
+            # find indexes of extrema not already found. Fraction index/degree
+            # cannot be further simplified
+            degree = int(2 ** (numpy.ceil(numpy.log2(i)) + 1))
+            if i == 1:
+                # special case for level == 1
+                indexes = numpy.arange(0, degree + 1, 2, dtype=int)
+            else:
+                indexes = indexes = numpy.arange(1, degree, 2, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, degree) > 1)]
```
mphoward: Please clean up this code to make it more readable like we did for the other nested CC class.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+            self._frequency = frequency
+            self._valid_cache = False
+
+    def _create(self):
+        r"""Create the points in the set.
+
+        Generating trigonometic points at a given frequency.
+        """
+        idx = numpy.arange(0, 2 * abs(self.frequency) + 1, 1, dtype=int)
+        points = (idx) * 2 * numpy.pi / (2 * abs(self.frequency) + 1)
```
mphoward: This is more readable & matches your documentation. I also would not cast the arange to an `int` because it needs to be converted to float again anyway.
```suggestion
        num_points = 2 * abs(self.frequency) + 1
        indexes = numpy.arange(num_points)
        points = 2 * numpy.pi *  indexes /  num_points
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        self.frequency = frequency
+
+    @property
+    def frequency(self):
+        """int: frequency to create points for."""
+        return self._frequency
+
+    @frequency.setter
+    def frequency(self, value):
+        frequency = int(value)
```
mphoward, line 421: Should we just document that `frequency` should be positive, and then either take the absolute value here to be sure or else error if it isn't? Then we don't have to do it below.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    r"""Set of unidimensional points for Trigonometric sampling
+
+    The :attr:`points` for this interpolation scheme are the
+    trigonometric interpolation points for m points
+
+    .. math::
+
+        x^f_j = \frac{2\pi j}{m(f)},  1 \leq j \leq m(l), f \geq 0
+
+    where f is the frequency. The relationship between the frequency
```
mphoward: You should clarify in the documentation what the meaning of this frequency is (i.e., what function do these points come from?)

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        r"""Create the points in the set.
+
+        Generating the trignometric points using the frequencies
+        :math:1, 3, 9, ..., 3^{i} where i is an integer.
+        """
+        # create properties for levels, level 0 is a special case with 1 point
+        rule = lambda x: 3**x
+        self._num_per_level = numpy.ones(self.num_level, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.num_level)
```
```suggestion
            rule(i) - rule(i - 1) for i in range(1, self.num_level)
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+        points = numpy.zeros(num_points, dtype=float)
+        for i in range(self.num_level):
+            # find fraction index/num_points that cannot be further simplified
+            num_points = 3**i
+            if i == 0:
+                # special case where level == 0
+                indexes = numpy.arange(0, num_points, dtype=int)
+            else:
+                indexes = numpy.arange(1, num_points + 1, 1, dtype=int)
+                indexes = indexes[~(numpy.gcd(indexes, num_points) > 1)]
```
mphoward: Clean up this code as above.

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+            points[self._start_level[i] : self._end_level[i]] = (
+                (indexes) * 2 * numpy.pi / num_points
```
```suggestion
                2 * numpy.pi * indexes / num_points
```

smolyay/samples.py
```diff
@@ -0,0 +1,496 @@
...
+    def _create(self):
+        r"""Create the points in the set.
+
+        Generating nested extrema of chebyshev polynomials of the first kind.
+        """
+        # create properties for levels, level 0 is a special case with 1 point
+        rule = lambda x: 1 if x == 0 else int(2 ** (numpy.ceil(numpy.log2(x)) + 1) + 1)
+        self._num_per_level = numpy.ones(self.num_level, dtype=int)
+        self._num_per_level[1:] = [
+            rule(i) - rule((i - 1)) for i in range(1, self.num_level)
```
```suggestion
            rule(i) - rule(i - 1) for i in range(1, self.num_level)
```

tests/samples_test.py
```diff
@@ -0,0 +1,416 @@
...
+    ),
+    (smolyay.samples.TrigonometricPointSet([0, 2 * numpy.pi], 0), [0]),
+    (
+        smolyay.samples.TrigonometricPointSet([0, 2 * numpy.pi], 1),
+        [0, 2 * numpy.pi / 3, 4 * numpy.pi / 3],
+    ),
+    (
+        smolyay.samples.TrigonometricPointSet([0, 2 * numpy.pi], -1),
+        [0, 2 * numpy.pi / 3, 4 * numpy.pi / 3],
+    ),
```
mphoward: If we force frequency to be positive (which I think makes sense), you can remove these negative ones from the test set, and just add a test that the right behavior occurs if you set the frequency to be a negative number (either error, or taking absolute value).

tests/samples_test.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+@pytest.mark.parametrize(
+    "samples,set_args",
+    [
+        (smolyay.samples.ClenshawCurtisPointSet, {"degree": 7}),
+        (smolyay.samples.TrigonometricPointSet, {"frequency": 7}),
+        (smolyay.samples.NestedClenshawCurtisPointSet, {"num_level": 7}),
+        (smolyay.samples.SlowNestedClenshawCurtisPointSet, {"num_level": 7}),
+        (smolyay.samples.NestedTrigonometricPointSet, {"num_level": 7}),
```
mphoward: Is it really necessary to use such a large degree / frequency / number of levels? I would think especially the large number of levels is not necessary, and might bog this test down if someone adds something in future that triggers the points to be created.

tests/samples_test.py
```diff
@@ -0,0 +1,416 @@
...
+        "NestedClenshawCurtis",
+        "SlowNestedClenshawCurtis",
+        "NestedTrigonometric",
+    ],
+)
+def test_num_level_error(nested_samples):
+    """test error given invalid num_level"""
+    with pytest.raises(ValueError):
+        f = nested_samples([-10, 10], 0)
+        f.domain = [[-10, 10], [-10, 10]]
```
mphoward: I think this is a copy paste error?
```suggestion
```

tests/samples_test.py
```diff
@@ -0,0 +1,416 @@
...
+    with pytest.raises(ValueError):
+        f = nested_samples([-10, 10], 2)
```
mphoward: This should go outside the test, so you make sure it constructs correctly and then only errors on set.
```suggestion
    f = nested_samples([-10, 10], 2)
    with pytest.raises(ValueError):
```

mphoward (approved):
Thanks Michaela! Please implement all the changes I have marked, and then this ready to merge from my side!

smolyay/samples.py
mzbush: It shouldn't error, since trigonometric polynomial frequencies are allowed to be negative. It can be forced to the absolute value.

mphoward: I am worried this behavior will be confusing because when the user enters a negative frequency, the object will give back a positive frequency.

For the purposes of generating a point set, I think it is a good idea to just require a positive frequency, especially now that we are generating the points in pairs. The negative frequency is more meaningful for the basis functions, and the user can always take the absolute value themselves if they want to work with a negative frequency. This is consistent with Python’s “explicit is better than implicit” concept.

Please require frequency to be nonnegative, and make this an error otherwise!

mzbush: Done

smolyay/samples.py
```diff
@@ -0,0 +1,510 @@
...
+    @property
+    def points(self):
+        """numpy.ndarray: Points in the set."""
+        if not self._valid_cache:
+            self._create()
+            self._valid_cache = True
+        return self._points
+
+    def __len__(self):
+        return len(self.points)
```
```suggestion
        return self.points.shape[0]
```

smolyay/samples.py
```diff
@@ -0,0 +1,510 @@
...
+    def end_level(self):
+        """numpy.ndarray: the ending index of each level."""
+        if not self._valid_cache:
+            self._create()
+            self._valid_cache = True
+        return self._end_level
+
+    def level(self, index):
+        """numpy.ndarray: Points in a level"""
+        return self[self.start_level[index] : self.end_level[index]]
```
```suggestion
        return self.points[self.start_level[index] : self.end_level[index]]
```

mphoward (approved):
Thanks! I had two minor suggestions that I will commit by the web interface, and then this is good to go. @cakieslich please review and merge when you get a chance!

cakieslich (approved):
Thanks! Looks good to me.

PR #22, opened 2024-04-23
Add optimum information

mzbush:
Add optimization information to the benchmark functions


smolyay/benchmark/beale.py
```diff
@@ -30,3 +30,11 @@ class beale(_beale):
...
         return ([[-7.0000000008, 11.69999999928],
                           [-9.5000000002,9.44999999982]])
+    
+    @property
+    def global_minimum(self):
+        return 0
+    
+    @property
+    def global_minimum_location(self):
+        return [2.9999999992, 0.4999999998]
```
mphoward: Is this how this came from Dr. K? I would assume this is actually (3, 0.5).

smolyay/benchmark/biggs.py
```diff
@@ -9,6 +9,14 @@ def domain(self):
                 [-1.96500000271271e-07, 17.99999982315],
                 [-5.0000001349, 13.49999987859]]
 
+    @property
+    def global_minimum(self):
+        return 0
+    
+    @property
+    def global_minimum_location(self):
+        return [0.9999999832, 9.9999998035, 4.9999998651]
```
mphoward, line 18: Same here. In general in these files, could you comment on where the really long decimals come from?

tests/benchmark_test.py
```diff
@@ -222,3 +232,9 @@ def test_functions_call_shape(fun):
     x_list[0] = numpy.linspace(fun.lower_bounds,mid_point,50)
     x_list[1] = numpy.linspace(mid_point,fun.upper_bounds,50)
     assert fun(x_list).shape == (2,50)
+
+@pytest.mark.parametrize("fun",functions)
+def test_functions_optimum(fun):
+    """Test all functions optimums are at their locations"""
+    print(fun.name)
```
mphoward:
```suggestion
```

mphoward (approved):
Thanks Michaela! I'm generally good with this, if you could just comment briefly on the question about where we think the weird precision values come from. Can you also summarize for us if there are files you did things other than just add the minima to? It's hard to see cause the diff is so big.

tests/benchmark_test.py
mzbush: Whenever this function errors or fails, it doesn't print out the name of the function, just the order in the list. I added a print so it's easier to determine what function needed to be investigated.

mzbush:
> Thanks Michaela! I'm generally good with this, if you could just comment briefly on the question about where we think the weird precision values come from. Can you also summarize for us if there are files you did things other than just add the minima to? It's hard to see cause the diff is so big.

1. benchmark.py - added abstract properties `global_minimum` and `global_minimum_location`
2. all files - added global minimum and global minimum location to all functions
3. denschn.py - reorganize functions so they are in alphabetical order
4. esfl.py - changed to using the gms code in the `_function` method
5. fermat.py - changed to using the gms code in the `_function` method for problems fermat_vareps and fermat2_vareps.
6. median.py -  changed to using the gms code in the `_function` method for median_vareps
7. palmer.py - changed to using the gms code in the `_function` method for problems palmer2a, palmer3, palmer4, and palmer8a.
8. yfit.py - changed to using the gms code in the `_function` method for yfit
9. growth.py - changed to using the gms code in the `_function` method for problems growth and growthls
10. least.py - added 20 to the first upper bound of least
11. himmel.py - changed bounds of himmelbf from [[0, 0.378], [0, 0.378], [0, 0.378], [0, 0.378]] to [[-10, 0.378], [0, 200], [0, 1800], [-50, 0.378]]
12. s.py - changes to s245, s266, s267, s351 that I will put in a separate list so it's not all one line.

Changes made to s.py
1. s245 - changed bounds from [[-9.0000000001, 0], [-7.90000065364893e-09, 0], [-8.9999999998, 0]] to [[-9.0000000001, 2], [-7.90000065364893e-09, 15], [-8.9999999998, 10]]
2. s266 - use gms code
3. s267 - increase upper bound of last dimension by 2
4. s351 - increase upper bound of 3rd dimension by 20

smolyay/benchmark/beale.py
mzbush: I was copied straight from the text file. This problem is easy to solve by hand, and I've confirmed that it is (3, 0.5) and I can go ahead and update it.

smolyay/benchmark/biggs.py
mzbush: According to @cakieslich, these were the best answers that the solvers found from wherever he got them. That may be the exact solution, or maybe the solver stopped short.

In that vein, I did some analysis to see if rounding decreased the error between the minimum recorded and one at the given location. Rounding to 0 decimal places improved the error for 31 problems, and of those 31, 24 problems got 0 error (and coincidentally all 24 had a min at 0). Changing the number of decimal places would change the number of improved problems slightly, so i want to do some more investigating to find the best min location.

 I'll do a commit changing the ones that got 0 error and see how that changes things.

mzbush: There were 5 functions where the rounding brought them farther from the recorded locations than usual, an around 1e-3 difference. These are himmelbb sisser, powell, s256, and s275. Looking at the functions written down, though, it's clear all of them have a global minima at 0 for all variables.

cakieslich:
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________
From: mzbush ***@***.***>
Sent: Friday, May 3, 2024 6:08:11 PM
To: biomodsquad/smolyak-surrogate-modeling ***@***.***>
Cc: Christopher Kieslich ***@***.***>; Mention ***@***.***>
Subject: [EXT] Re: [biomodsquad/smolyak-surrogate-modeling] Add optimum information (PR #22)

CAUTION: Email Originated Outside of Auburn.

@mzbush commented on this pull request.

________________________________

In smolyay/benchmark/biggs.py<https://github.com/biomodsquad/smolyak-surrogate-modeling/pull/22#discussion_r1589760800>:

> @@ -9,6 +9,14 @@ def domain(self):
                 [-1.96500000271271e-07, 17.99999982315],
                 [-5.0000001349, 13.49999987859]]

+    @property
+    def global_minimum(self):
+        return 0
+
+    @property
+    def global_minimum_location(self):
+        return [0.9999999832, 9.9999998035, 4.9999998651]


There were 5 functions where the rounding brought them farther from the recorded locations than usual, an around 1e-3 difference. These are himmelbb sisser, powell, s256, and s275. Looking at the functions written down, though, it's clear all of them have a global minima at 0 for all variables.

mphoward:
What work (if any) needs to be done to finalize this PR?

mphoward:
@mzbush did we decide that this PR should be closed?

mphoward:
I will leave this branch around for a little bit if we decide it should be retargeted onto `main`.

PR #23, opened 2024-05-20
Add normalizers

mzbush:
Implements a class of objects that can apply a normalizing transformation to the surrogate's training data and the inverse transformation to its output.

The class is constructed in a way to make it compatible with certain properties of the sklearn.preprocessing transformation classes. However, these objects do not assume there are multiple features to the data and are more flexible in the shape of inputs that are allowed.


mphoward:
We discussed this offline, and we are not going to pursue implementing normalization. The user can do this themselves pre-training.

PR #24, opened 2024-06-11
Multidimensional Point Set

mzbush:
The multidimensional point sets


smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
-        """:class:`BasisFunctionSet`: basis set (points and functions)."""
-        return self._basis_set
-
-    @basis_set.setter
-    def basis_set(self, basis_set):
-        if not isinstance(basis_set, BasisFunctionSet):
-            raise TypeError('Basis set must be a BasisFunctionSet')
-        self._basis_set = basis_set
+    @abc.abstractmethod
+    def dimension(self):
```
mphoward: Should this be:
```suggestion
    def num_dimensions(self):
```
?

Also, this can be inferred from the shape of the domain, so it does not need to be abstract.
```
return self.domain.shape[0]
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
-        """Make grid points and their corresponding basis functions.
+    def domain(self):
+        """numpy.ndarray: domain of the point set."""
+        pass
+
+    @property
+    def points(self):
+        """numpy.ndarray: Points to be sampled."""
+        if self._points is None:
+            self._generate_points()
```
mphoward: This should follow the same naming convention and cache pattern as the `UnidimensionalPointSet`

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+        return self._points
+
+    @staticmethod
+    def _scale_to_domain(x, old, new):
```
mphoward: This should follow the same convention as the unidimensional point set.

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+        else:
+            for i in range(new.shape[0]):
+                numpy.clip(new_x[..., i], new[i, 0], new[i, 1], out=new_x[..., i])
+        return new_x
+
+    def set_params(self, **parameters):
+        for parameter, value in parameters.items():
+            setattr(self, parameter, value)
+        self._generate_points()
+        return self
```
mphoward: It is better to have actual attributes (as properties), rather than use a catch all method like this.
```suggestion
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
 
+    options: {None, dict}, optional
+        if method != None, additional parameters passed to the QMCEngine
     """
 
-    def __init__(self, basis_set):
-        super().__init__(basis_set)
-        if not isinstance(self.basis_set, NestedBasisFunctionSet):
-            raise TypeError('Basis set must be nested for Smolyak grid')
+    def __init__(self, domain, number_points, method, seed, options=None):
```
mphoward: Each of these `method` options should be its own class, that takes its options as explicit constructor arguments / makes them available as properties. You should still use a base `RandomPointSet` that supplies the common parts for all of them. For example,

```py
class UniformRandomPointSet(RandomPointSet):
    ...
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
-        their corresponding basis function and levels are then
-        generated by replacing the indexes with their
-        corresponding unidimensional points, functions, and levels' indexes.
+    @property
+    def domain(self):
+        """numpy.ndarray: domain of the point set."""
+        if self.dimension > 1:
+            return numpy.array(self._domain)
+        else:
+            return numpy.array(self._domain[0])
```
mphoward: If these are multidimensional point sets, I don't think we should flatten the shape of the domain. You can make unidimensional point sets if you want to have access to random point sets that behave like this.

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+
+    @domain.setter
```
mphoward: Shouldn't this stuff be in the base `MultidimensionalPointSet` class, following the pattern of the unidimensional point sets?

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
-    Depending on the dimensionality, points and basis functions provided
-    by :class:`BasisFunctionSet`, the :meth:`__call__` makes full tensor
-    grids.
-    :meth:`__call__`  generates the indexes of
-    grid points (integer grids) depending on dimensionality (full tensor grid).
-    Grid points and their functions are made by replacing the indexes with
-    corresponding point.
-    """
+    @property
+    def number_points(self):
```
mphoward: This would be a length method, following what you did for the unidimensional point sets and belong in the base class.
```suggestion
    def __len__(self):
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+
+        Using the parameters of the class, generate a grid set using
+        the a Monte Carlo/Quasi Monte Carlo method.
+        """
+        self._points = self.get_random_points(
+            self.domain, self.number_points, self.method, self.seed, self.options
+        )
+
+    @staticmethod
+    def get_random_points(domain, number_points, method, seed, options=None):
```
mphoward: This is a good candidate to turn into an abstract method that returns the generator for a specific type of random point set, then you can have a common create points method that uses that generator.

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+        return len(self.point_sets)
+
+    @property
+    def point_sets(self):
+        """:class:UnidimensionalPointSet: set of unique points"""
+        return self._point_sets
+
+    @point_sets.setter
+    def point_sets(self, value):
+        self._point_sets = value
```
mphoward: We discussed not exposing these point sets since we don't want them to be mutable for now.

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
-    @property
-    def points(self):
-        """list: Grid points."""
-        return self._points
 
-    @property
-    def basis_functions(self):
-        """list: Basis functions of the grid points."""
-        return self._basis
+class TensorPointSet(PointSetProduct):
```
mphoward:
```suggestion
class TensorProductPointSet(PointSetProduct):
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        self._points = numpy.array(list(itertools.product(*self._point_sets)))
+        
 
 
-class NestedIndexGrid(IndexGrid):
-    """Set of grid points and their corresponding functions (a data structure).
+class SmolyakPointSet(PointSetProduct):
```
mphoward: Maybe try to include sparse?
```suggestion
class SmolyakSparseProductPointSet(PointSetProduct):
```

smolyay/grid.py
```diff
@@ -2,348 +2,429 @@
...
+        """Generating the tensor grid
 
-    @property
-    def levels(self):
-        """list: level each grid points' basis function was taken from."""
-        return self._levels
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        grid_points = None
```
mphoward: This code probably needs cleaning up, but I'm not going to review right now.

smolyay/grid.py
mphoward, line 1: I think all of these belong in the `samples.py` because they are also types of samples. I don't think there is a need for a `grid.py`.

mphoward (changes requested):

smolyay/grid.py
mzbush: It's always been referred to as "dimension" by the surrogate class and the grid generators.

mphoward: Yes, but we are **refactoring** this code, so things should be changed to be consistent. In this case, we prefix things with `num_` when they refer to a number of something.

smolyay/grid.py
mzbush: The PointSetProduct gets its domain from the UnidimensionalPointSet, so it doesn't have its own setter.

smolyay/grid.py
mzbush: Should it even exist if the unidimensional point set already takes care of scaling the domain?

mphoward: If the random point sets don't need this, then yeah, you should probably remove this method.

smolyay/grid.py
mphoward: If you want to start to allow a basic level of mutability of the underlying unidimensional point sets, you could forward changes in the domain to the underlying unidimensional point sets.

At minimum, I think you can put the read-only property in the base class, and then add the setter in children that allow one.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
                 2 * numpy.pi * indexes / num_points
             )
         self._points = self._scale_to_domain(points, [0, 2 * numpy.pi])
+
+
+import abc
+import itertools
+
+import numpy
+import scipy.stats.qmc
```
mphoward: Imports must go at the top of the file.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+
+    def __init__(self):
+        self._domain = None
+        self._points = None
+        self._valid_cache = False
+
+    @property
+    def num_dimensions(self):
+        """int: number of independent variables."""
+        return numpy.shape(self.domain)[0]
```
mphoward: `domain` should be a numpy array, so this will work
```suggestion
        return self.domain.shape[0]
```

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        return self.points.shape[0]
+
+    @abc.abstractmethod
+    def _create(self):
+        """Abstract method for generating the mulitdimensional grid set"""
+        pass
+
+
+class RandomPointSet(MultidimensionalPointSet):
+    """Point Set that uses randomly generated points
```
mphoward: This documentation is out of date.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        self._seed = None
+
+        self.domain = domain
+        self.number_points = number_points
+        self.seed = seed
+
+    @property
+    def domain(self):
+        """numpy.ndarray: domain of the point set."""
+        return self._domain
```
mphoward: Shouldn't this be inherited from the parent? Does that not work for properties?

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+    seed: {int, numpy.random.Generator}
+        seed for generating the random points
+
+    options: {None, dict}, optional
+        if method != None, additional parameters passed to the QMCEngine
+    """
+
+    def __init__(self, domain, number_points, seed):
+        super().__init__()
+        self._domain = None
```
mphoward: This is inherited, yes?
```suggestion
```

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        if domain.ndim != 2 or domain.shape[1] != 2:
+            raise TypeError("Domain must have size (num_dimensions, 2)")
+        if any(domain[:, 0] >= domain[:, 1]):
+            raise ValueError("Lower bound must be less than upper bound")
+        if not numpy.array_equal(self._domain, domain):
+            self._domain = domain
+            self._valid_cache = False
+
+    @property
+    def number_points(self):
```
mphoward: This would be `num_points` with our naming convention.
```suggestion
    def num_points(self):
```

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+            self._valid_cache = False
+
+    @property
+    def seed(self):
+        """int: random seed for generating points"""
+        return self._seed
+
+    @seed.setter
+    def seed(self, value):
+        if not isinstance(value, numpy.random.Generator):
```
mphoward: Why do we need to support `numpy.random.Generator` and not just take a seed value?

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+
+        Generates a set of random points with a given Monte Carlo/Quasi Monte
+        Carlo method depending on the class.
+
+        Returns
+        -------
+        numpy.ndarray
+            generated random points
+        """
+        pass
```
mphoward: I think you don't need this pass through here. Just implement `_create` for each subclass.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+                ps.domain = d
+            self._valid_cache = False
+
+    @property
+    def point_sets(self):
+        """:class:UnidimensionalPointSet: set of unique points"""
+        return self._point_sets
+
+    @property
+    def indexes(self):
```
mphoward: I don't think this is part of the public API we discussed. This point set should expose the points (not indexes).

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        unique set of 1D points
+    """
+
+    def _create(self):
+        """Generating the tensor grid
+
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        self._points = numpy.array(list(itertools.product(*self._point_sets)))
```
mphoward: This is going to cause a double memory allocation since it makes a list then an array. I think something like this would be good:
```suggestion
        num_points = numpy.prod([len(p) for p in self._point_sets])
        self._points = numpy.array((num_points, self.num_dimensions), dtype=float)
        for i, point in enumerate(itertools.product(*self._point_sets)):
            self._points[i] = point
```

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+
+    def _create(self):
+        """Generating the tensor grid
+
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        grid_points = None
+        max_num_level = max([ob.num_levels for ob in self._point_sets])
+        max_num_levels = [ob.num_levels for ob in self._point_sets]
```
```suggestion
        max_num_levels = [ob.num_levels for ob in self._point_sets]
        max_num_level = max(max_num_levels)
```
It would be good if you can choose a better name for the second variable, it will get confusing below.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        # get the combinations of levels
+        index_composition = []
+        for sum_of_levels in range(
+            self.num_dimensions, self.num_dimensions + max_num_level
```
mphoward: Given that our levels are zero-indexed in Python, can we change this to be a `range(max_num_level)`, include zero below, and not worry about all the shifting?

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+                    )
+                )
+            )
+        index_composition = numpy.array(index_composition) - 1
+        if min(max_num_levels) != max_num_level:
+            # only check if point sets have different number of levels
+            valid_comb = numpy.all(
+                numpy.greater_equal(max_num_levels, index_composition), axis=1
+            )
+            index_composition = index_composition[valid_comb, :]
```
mphoward: What is this doing? Some commenting, etc. would be helpful.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+            level_composition_index = [
+                self._point_sets[d].level(index) for d, index in enumerate(index_comp)
+            ]
+            grid_points_ = numpy.array(
+                numpy.meshgrid(*level_composition_index)
+            ).T.reshape(-1, self.num_dimensions)
+            if grid_points is None:
+                grid_points = grid_points_
+            else:
+                grid_points = numpy.concatenate((grid_points, grid_points_), axis=0)
```
mphoward: Similarly, some commenting here would be helpful. I think this is taking the tensor products of the relevant levels?

I would in general look at whether you can rename some variables here based on our new notation to make things clearer. For example, `grid_points` probably is not meaningful and should just be `points`. Similarly, `index_composition`, `level_composition_index`, etc. might be named more clearly.

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+        lower_bounds = [bound[0] for bound in self.domain]
+        upper_bounds = [bound[1] for bound in self.domain]
+        num_dimensions = len(lower_bounds)
+        p_gen = numpy.random.default_rng(seed=self.seed).uniform(
+            size=(self.number_points, num_dimensions)
+        )
+        return scipy.stats.qmc.scale(p_gen, lower_bounds, upper_bounds)
+
+
+class QMCRandomPointSet(RandomPointSet):
```
mphoward: I don't think this class should be part of the public API. It is just a shim between `RandomPointSet` and the implementations below.
```suggestion
class _QMCRandomPointSet(RandomPointSet):
```

smolyay/samples.py
```diff
@@ -508,3 +508,620 @@ def _create(self):
...
+    @property
+    def bits(self):
+        """{1, 2}: Whether to create an orthogonal array of points"""
+        return self._bits
+
+    @bits.setter
+    def bits(self, value):
+        bits = int(value)
+        if bits > 64:
+            raise ValueError("bits max value is 64.")
```
mphoward: You need to check for a positive power of 2, right? This is the bitwise math to do that.
```suggestion
        if bits > 64:
            raise ValueError("bits max value is 64.")
        elif bits < 1:
            raise ValueError("bits must be at least 1")
        elif bits & (bits-1) != 0:
            raise ValueError("bits must be a power of 2")
```

mphoward (changes requested):
Thanks Michaela! Here are some more comments on the code. We can look at the tests later.

smolyay/samples.py
mzbush: For some reason, if I delete this getter method it stops working. It seems like if the child class has **only** the setter method for a property it should inherit from a parent class, it doesn't count somehow. I'm not sure if this is a consequence of the parent class not having a setter method for the child to override or what.

smolyay/samples.py
mzbush: I only supported it because numpy.random.default_rng and the QMCEngines can take either. I viewed these classes as merely containers for initializing the QMCEngine and calling `numpy.random.default_rng(seed=self.seed).uniform()`, so in my mind it made sense for it to be as flexible with parameter typing as the QMCEngine and numpy.random. 

I can force it to only take ints if you prefer that.

smolyay/samples.py
mzbush: `bits` doesn't have to be a power of 2. It's default value is 30, after all. `bits` is related to the maximum number of points before the Sobol sequence repeats itself which the QMCEngine doesn't allow, and that max number is `2**bits`.

mphoward: Oh, I misread the errors above that about bits and points. You can remove this check on power of 2 here, but this is how you should do it for `num_points` below. It is exact because it is bitwise integer math.

smolyay/samples.py
mphoward: Let's just use an int, it's easier to understand.

smolyay/samples.py
mphoward: Gotcha. SO suggests you include the name of the parent class when you define a setter in a child, like:
```
@MultidimensionalPointSet.domain.setter
def domain(self, value):
    ...
```
If that doesn't work, I think you could do the getter like:
```
@property
def domain(self):
    return super().domain
```
to at least avoid duplicating its implementation.

https://stackoverflow.com/questions/57230274/property-setter-and-inheritance

smolyay/samples.py
mzbush: Yes we can.

smolyay/samples.py
mzbush: Basically, when generating the combinations of levels for each dimension, it generates them first assuming that all dimensions can take the same number of levels. This part of the code first checks if that's actually true, and if it isn't, it finds all the valid combinations and only keeps those. Valid combinations are ones where the level for every dimension doesn't exceed the number of levels for that dimension.

I have added comments and cleaned up this section of code.

smolyay/samples.py
mzbush: Yeah. For each combination of levels, it grabs the list of points associated with the level number and then uses meshgrid to make the tensor products.

smolyay/samples.py
mzbush: What about doing something like this instead?
It worked when I tested it, and the runtime doesn't grow as fast.
`self._points = numpy.array(numpy.meshgrid(*self._point_sets)).T.reshape(-1, self.num_dimensions)`

mzbush: Using self.num_dimensions by necessity creates a dependency on the self.domain property, which for PointSetProduct creates a dependency on the elements in self._point_sets having a domain property.

Overridding the getter for num_dimensions in the PointSetProduct to `return len(self.point_sets)` removes that dependency. You could put in a list of numpy arrays as the point_sets and the tensor would be able to generate grid points just fine. The UnidimensionalPointSets can be used like lists, and while that's not always the case in reverse, it would be very trivial to make the tensor capable of using pure lists in place of unidimensional point sets.

mphoward: You could add `num_dimensions = len(self._point_sets)` in my code above instead of `self.num_dimensions`, rather than need to touch aany other code. Unless there is a drastic time bottleneck, I think you are better off using the lower memory version I suggested. Your array and meshgrid will probably trigger at least 1 extra allocation, and this is bad for large tensor products.

mzbush: What about 

`self._points = numpy.stack(numpy.meshgrid(*self._point_sets), -1).reshape(
            -1, len(self._point_sets))`

It outputs the points in a different order but it ended up being slightly faster in the higher dimensions than the first meshgrid idea.

mphoward: Is there a reason you can’t use the code as I wrote it? It uses an iterator so the array does not get generated twice. I am pretty sure yours will due to the combination of meshgrid with other array operations. Memory considerations are very important when arrays can get large, which is the case for the tensor product.

mzbush: Simply that using the iterator has a worse runtime complexity than either of the meshgrid strategies. If there was some strategy and combination of operations that allowed the best of both worlds, the meshgrid strategies runtime complexity without significant memory allocations, then I would want that one. I am just trying to see if I can find a strategy that doesn't make extra copies of the array.

As it stands, we can use the iterator strategy.

mphoward: Do you have a timing that shows how much slower the iterator is? In general, you shouldn’t optimize until you profile and find something is a bottleneck. In this case, I don’t really care if this bit of code is a little slower because it will typically be done only once and so will probably be a negligible cost most of the time. I do care, though, if the calculation gets killed that one time it runs because Python ran out of memory.

Just FYI: complexity and time aren’t the same thing. The complexity of both approaches is O(N), I believe, but one is just slower because it isn’t vectorized by NumPy.

mzbush: I do. I should be able to show it tomorrow.

mzbush: I did 2 test runs with all the different methods (yours with itertools, using meshgrid and transpose, and using meshgrid and stack).
image

This was the runtimes for all three methods making a 2D grid, a 3D grid, and a 4D grid. As the number of unique values used to make the grid increases, the itertools methods takes much longer and the time difference gets more drastic as the dimensions increase. I made the y axis logarithmic so the runtimes making 2D grids is visible.
image

This is a different run that includes 5D and 6D grids. The intertools method making an n dimensional grid tends to have a runtime comparable to making a n+1 dimensional grid using the 2 meshgrid methods.

In retrospect, the time differences aren't that big a deal since we're going to be mostly working on a small number of unique points, (the largest 6D grid uses only 5 unique points), but it's a time save that would be nice to have if possible.

mphoward: Given that these differences probably aren't significant to the overall problem (solving the coefficients of a 15^6 grid likely takes way longer than 1 s), we should use the `itertools` version for readability and to reduce memory usage.

Moving forward, I recommend that you don't worry about running microbenchmarks like this for now. It will be faster to get a working prototype (coded however we think is best), then profile actual use cases to see where performance optimization is needed.

You may find it interesting to read about [Amdahl's law](https://en.wikipedia.org/wiki/Amdahl%27s_law).

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+class MultidimensionalPointSet(abc.ABC):
+    """Multidimensional set of points
+
+    Creates a set of grid points within the specified domain
+
+    Parameters
+    ----------
+    domain: list
+         the domain of the grid
+    """
```
```suggestion
    """Multidimensional set of points."""
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        self.domain = domain
+        self.num_points = num_points
+        self.seed = seed
+
+    @MultidimensionalPointSet.domain.setter
+    def domain(self, value):
+        domain = numpy.sort(numpy.array(value, ndmin=2), axis=1)
+        if domain.ndim != 2 or domain.shape[1] != 2:
+            raise TypeError("Domain must have size (num_dimensions, 2)")
+        if any(domain[:, 0] >= domain[:, 1]):
```
```suggestion
        if numpy.any(domain[:, 0] >= domain[:, 1]):
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            raise ValueError("Lower bound must be less than upper bound")
+        if not numpy.array_equal(self._domain, domain):
+            self._domain = domain
+            self._valid_cache = False
+
+    @property
+    def num_points(self):
+        """int: random seed for generating points"""
```
```suggestion
        """int: Number of points"""
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    @seed.setter
+    def seed(self, value):
+        value = int(value)
+        if self._seed != value:
+            self._seed = value
+            self._valid_cache = False
+
+
+class UniformRandomPointSet(RandomPointSet):
+    """Generates a grid using a uniform distribution"""
```
```suggestion
    """Points drawn from a uniform random distribution."""
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+
+    def _create(self):
+        """Generate a set of random points
+
+        Generates uniformly distributed random points
+        """
+        lower_bounds = [bound[0] for bound in self.domain]
+        upper_bounds = [bound[1] for bound in self.domain]
```
mphoward: These are NumPy arrays now, so you should be able to slice, right?
```suggestion
        lower_bounds = self.domain[:, 0]
        upper_bounds = self.domain[:, 1]
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    strength : {1, 2}, optional
+        Default 1. If 1, produces a normal LHS. If 2, produces an orthogonal
+        array based LHS.
+    """
+
+    def __init__(
+        self, domain, num_points, seed, scramble=True, optimization=None, strength=1
+    ):
+        super().__init__(domain, num_points, seed, scramble, optimization)
+        self._strength = 1
```
mphoward: Is there a reason not to assign this as None initially? The behavior is slightly an anti-pattern.
```suggestion
        self._strength = None
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        if self._strength != strength:
+            self._strength = strength
+            self._valid_cache = False
+
+    def _create(self):
+        """Generate a set of random points
+
+        Generates a set of monte carlo LatinHypercube points.
+        """
+        lower_bounds = [bound[0] for bound in self.domain]
```
mphoward: Slice as above.

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        Default 30. Sets the max number of points that can be generated,
+        which is 2**bits.
+    """
+
+    def _create(self):
+        """Generate a set of random points
+
+        Generates a set of quasi-monte carlo Halton Sequence points.
+        """
+        lower_bounds = [bound[0] for bound in self.domain]
```
mphoward: Slice as above

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    domain: list
+        domain of the random points
+
+    num_points: int
+        number of points to generate
+
+    seed: int
+        seed for generating the random points
+
+    scramble : bool
```
mphoward, line 778: Fix this doc.

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        scrambling to Halton points, and LMS+shift scrambling to Sobol points.
+
+    optimization : {None, "random-cd", "lloyd"}
+        Default None. If "random-cd" the coordinates of points are adjusted to
+        lower the centered discrepancy. If "lloyd", adjust points using a
+        Lloyd-Max algorithm to encourage even spacing.
+
+    bits : int
+        Default 30. Sets the max number of points that can be generated,
+        which is 2**bits.
```
```suggestion
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    domain: list
+        domain of the random points
+
+    num_points: int
+        number of points to generate. Must be a power of 2.
+
+    seed: int
+        seed for generating the random points
+
+    scramble : bool
```
mphoward, line 818: Fix this doc.

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+
+    @bits.setter
+    def bits(self, value):
+        bits = int(value)
+        if bits > 64:
+            raise ValueError("bits max value is 64")
+        elif bits < 1:
+            raise ValueError("bits must be at least 1")
+        if 2**bits < self.num_points:
+            raise ValueError("2**bits must be greater than number of points")
```
mphoward: It isn't required, but I like doing all error checking related to a value in `if-elif` so I remember they are connected.
```suggestion
        elif 2**bits < self.num_points:
            raise ValueError("2**bits must be greater than number of points")

```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    def __init__(
+        self, domain, num_points, seed, scramble=True, optimization=None, bits=30
+    ):
+        self._bits = 64  # max value of bits
+        super().__init__(domain, num_points, seed, scramble, optimization)
+        self.bits = bits
+
+    @property
+    def bits(self):
+        """{1, 2}: Whether to create an orthogonal array of points"""
```
mphoward: This doc is incorrect

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        number of points must be a power of two
+    ValueError
+        number of points must be less than 2**bits
+    """
+
+    def __init__(
+        self, domain, num_points, seed, scramble=True, optimization=None, bits=30
+    ):
+        self._bits = 64  # max value of bits
+        super().__init__(domain, num_points, seed, scramble, optimization)
```
mphoward: This is an anti-pattern here. Call super first unless there is a reason not to, and then I prefer `None` if you are going to overwrite immediately anyway.
```suggestion
        super().__init__(domain, num_points, seed, scramble, optimization)
        self._bits = None
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    @num_points.setter
+    def num_points(self, value):
+        num_points = int(value)
+        if num_points & (num_points - 1) != 0:
+            raise ValueError("Number of points must be power of 2")
+        if num_points > 2**self.bits:
+            raise ValueError("Number of points must be less than 2**bits")
+        if self._num_points != num_points:
+            self._num_points = num_points
+
```
```suggestion
        elif num_points > 2**self.bits:
            raise ValueError("Number of points must be less than 2**bits")

        if self._num_points != num_points:
            self._num_points = num_points
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            self._valid_cache = False
+
+    def _create(self):
+        """Generate a set of random points
+
+        Generates a set of quasi-monte carlo Sobol Sequence points.
+        """
+        lower_bounds = [bound[0] for bound in self.domain]
```
mphoward: Slice

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    def domain(self):
+        """numpy.ndarray: domain of the point set."""
+        return numpy.array([up.domain for up in self.point_sets])
+
+    @domain.setter
+    def domain(self, value):
+        domain = numpy.sort(numpy.array(value, ndmin=2), axis=1)
+        if domain.ndim != 2 or domain.shape[1] != 2:
+            raise TypeError("Domain must have size (num_dimensions, 2)")
+        if any(domain[:, 0] >= domain[:, 1]):
```
```suggestion
        if numpy.any(domain[:, 0] >= domain[:, 1]):
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            self.num_dimensions,
+            scramble=self.scramble,
+            optimization=self.optimization,
+            seed=self.seed,
+        ).random(n=self.num_points)
+        self._points = scipy.stats.qmc.scale(p_gen, lower_bounds, upper_bounds)
+
+
+class SobolRandomPointSet(_QMCRandomPointSet):
+    """Generates a grid using Sobol Sequence
```
```suggestion
    """Points drawn randomly using a Sobol sequence.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            scramble=self.scramble,
+            strength=self.strength,
+            optimization=self.optimization,
+            seed=self.seed,
+        ).random(n=self.num_points)
+        self._points = scipy.stats.qmc.scale(p_gen, lower_bounds, upper_bounds)
+
+
+class HaltonRandomPointSet(_QMCRandomPointSet):
+    """Generates a grid using Halton Sequences
```
```suggestion
    """Points drawn randomly using a Halton sequence.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+    def optimization(self, value):
+        if not value in [None, "random-cd", "lloyd"]:
+            raise TypeError("optimization must be None, random-cd, or lloyd")
+        if self._optimization != value:
+            self._optimization = value
+            self._valid_cache = False
+
+
+class LatinHypercubeRandomPointSet(_QMCRandomPointSet):
+    """Generates a grid using a LatinHypercube
```
```suggestion
    """Points drawn randomly using a Latin hypercube.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            scramble=self.scramble,
+            bits=self.bits,
+            optimization=self.optimization,
+            seed=self.seed,
+        ).random(n=self.num_points)
+        self._points = scipy.stats.qmc.scale(p_gen, lower_bounds, upper_bounds)
+
+
+class PointSetProduct(MultidimensionalPointSet):
+    """Generate a grid using combinations of a unidimensional point set
```
```suggestion
    """Points generated from combinations of unidimensional points.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            self._valid_cache = False
+
+    @property
+    def point_sets(self):
+        """:class:UnidimensionalPointSet: set of unique points"""
+        return self._point_sets
+
+
+class TensorProductPointSet(PointSetProduct):
+    """Generate a grid using all combinations of unidimensional point sets
```
```suggestion
    """Points generated using all combinations of unidimensional points.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        num_points = numpy.prod([len(p) for p in self._point_sets])
+        self._points = numpy.zeros((num_points, len(self._point_sets)), dtype=float)
+        for i, point in enumerate(itertools.product(*self._point_sets)):
+            self._points[i] = point
+
+class SmolyakSparseProductPointSet(PointSetProduct):
+    """Generate a grid using sparse combinations of unidimensional point sets
```
```suggestion
    """Points generated from sparse combinations of unidimensional points.
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+
+    Depending on the dimensionality, points and provided by
```
mphoward: Fix this doc string. (In general, please review all doc strings of classes.)

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+
+    def _create(self):
+        """Generating the tensor grid
+
+        Using the parameters of the class, generate a grid set using all
+        combinations of unidimensional points
+        """
+        # reset points
+        self._points = None
+        # find number of levels per dimension
```
mphoward, line 983: Please add some whitespace to this function to make it easier to read. Probably a line before each of your comments would make sense.
```suggestion

        # find number of levels per dimension
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+                )
+            )
+        level_combinations = numpy.array(level_combinations)
+        # remove combinations where a dimension exceeds its number of levels
+        # only check if point sets have different numbers of levels
+        if min(num_levels_per_dim) != max_num_levels:
+            valid_comb = numpy.all(
+                numpy.greater(num_levels_per_dim, level_combinations), axis=1
+            )
+            level_combinations = level_combinations[valid_comb, :]
```
mphoward: The colon shouldn't be needed here
```suggestion
            level_combinations = level_combinations[valid_comb]
```

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+                        sum_of_levels, self.num_dimensions, include_zero=True
+                    )
+                )
+            )
+        level_combinations = numpy.array(level_combinations)
+        # remove combinations where a dimension exceeds its number of levels
+        # only check if point sets have different numbers of levels
+        if min(num_levels_per_dim) != max_num_levels:
+            valid_comb = numpy.all(
+                numpy.greater(num_levels_per_dim, level_combinations), axis=1
```
mphoward: Can we rewrite this to use `less`? I got confused reading this because the variable you care about is second. Also, double check here that it's implemented correctly since the level combinations are now zero-indexed. I think it should be:
```py
valid_comb = numpy.all(numpy.less(level_combinations, num_levels_per_dim), axis=1)
```
?

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+            level_point_combinations = [
+                self._point_sets[d].level(level) for d, level in enumerate(level_comb)
+            ]
+            points_ = numpy.array(numpy.meshgrid(*level_point_combinations)).T.reshape(
+                -1, self.num_dimensions
+            )
+            if self._points is None:
+                self._points = points_
+            else:
+                self._points = numpy.concatenate((self._points, points_), axis=0)
```
mphoward: Can we replace this with `itertools.product` as well to avoid large memory needs?

smolyay/samples.py
```diff
@@ -508,3 +511,583 @@ def _create(self):
...
+
+
+def generate_compositions(value, num_parts, include_zero):
```
mphoward: Make this private with a leading underscore, I don't think we want to guarantee it as part of the API.
```suggestion
def _generate_compositions(value, num_parts, include_zero):
```

mphoward (changes requested):

smolyay/samples.py
mzbush: The super calls the number of points setter which checks if the number of points exceeds bits. This means that bits needs to be set to something before the number of points setter is called, so here it's set to the max to ensure that the "Number of points must be less than 2**bits" error doesn't occur.

smolyay/samples.py
mzbush: Either should work. There is a test for different numbers of levels in each dimension.

mphoward: Please rewrite to use less, as it is more readable

smolyay/samples.py
mphoward: I see, please add a more explicit comment about why this is coming in this order then

smolyay/samples.py
mzbush: Done.

smolyay/samples.py
mzbush: I did. It should be ready for review again now.

smolyay/samples.py
```diff
@@ -508,3 +511,580 @@ def _create(self):
...
+
+    optimization : {None, "random-cd", "lloyd"}
+        Default None. If "random-cd" the coordinates of points are adjusted to
+        lower the centered discrepancy. If "lloyd", adjust points using a
+        Lloyd-Max algorithm to encourage even spacing.
+    """
+
+    def __init__(self, domain, num_points, seed, scramble=True, optimization=None):
+        super().__init__(domain, num_points, seed)
+        self._scramble = True
```
mphoward: Default to `None` to force the setter to run?
```suggestion
        self._scramble = None
```

smolyay/samples.py
```diff
@@ -508,3 +511,580 @@ def _create(self):
...
+        """bool: If True, applies centering to Latin Hypercube points, Owen
+        scrambling to Halton points, and LMS+shift scrambling to Sobol points."""
+        return self._scramble
+
+    @scramble.setter
+    def scramble(self, value):
+        scramble = bool(value)
+        if self._scramble != scramble:
+            self._scramble = scramble
+            self._scramble = False
```
```suggestion
            self._valid_cache = False
```

smolyay/samples.py
```diff
@@ -508,3 +511,580 @@ def _create(self):
...
+    Parameters
+    ----------
+    point_sets : list of :class:UnidimensionalPointSet
+        unique set of 1D points
+    """
+
+    def __init__(self, point_sets):
+        super().__init__()
+        self._point_sets = point_sets
+        self._indexes = []
```
mphoward: This is unused, right?
```suggestion
```

smolyay/samples.py
```diff
@@ -508,3 +511,580 @@ def _create(self):
...
+            valid_comb = numpy.all(numpy.less(level_combinations, num_levels_per_dim), axis=1)
+            level_combinations = level_combinations[valid_comb]
+
+        # generate sets of points based on combinations of levels
+        for level_comb in level_combinations:
+            level_point_combinations = [
+                self._point_sets[d].level(level) for d, level in enumerate(level_comb)
+            ]
+            num_points = numpy.prod([len(p) for p in level_point_combinations])
+            points_ = numpy.zeros((num_points, len(level_point_combinations)), dtype=float)
```
mphoward: Can you use `self.num_dimensions` here?
```suggestion
            points_ = numpy.zeros((num_points, self.num_dimensions), dtype=float)
```

tests/grid_test.py
```diff
@@ -1,126 +1,357 @@
...
+def test_random_initalize(random_point_set):
+    """Test that the random point set initializes correctly"""
+    f = random_point_set([[-10, 10], [0, 2]], 64, 1234)
+    assert numpy.array_equal(f.domain, [[-10, 10], [0, 2]])
+    assert f.num_dimensions == 2
+    assert f.num_points == 64
+    assert isinstance(f.num_points, int)
+    assert f.seed == 1234
+    assert isinstance(f.seed, int)
+    f.domain = [-10, 10]
```
mphoward, line 29: Make sure you use white space to make things readable. I almost missed this was testing both 2d and 1d domains & the setters
```suggestion

    f.domain = [-10, 10]
```

tests/grid_test.py
```diff
@@ -1,126 +1,357 @@
...
+    """Test that the random point set initializes correctly"""
+    f = qmc_point_set([[-10, 20]], 64, 5678, True, "random-cd")
+    assert numpy.array_equal(f.domain, [[-10, 20]])
+    assert f.num_dimensions == 1
+    assert f.num_points == 64
+    assert f.seed == 5678
+    assert isinstance(f.scramble, bool)
+    assert f.scramble == True
+    assert f.optimization == "random-cd"
+    f.scramble = 0
```
mphoward, line 59: Same here
```suggestion

    f.scramble = 0
```

tests/grid_test.py
```diff
@@ -1,126 +1,357 @@
...
+    assert isinstance(f.scramble, bool)
+    assert f.scramble == False
```
mphoward: I think Python prefers this:
```suggestion
    assert f.scramble is False
```
otherwise, use `assert not f.scramble`

tests/grid_test.py
```diff
@@ -1,126 +1,357 @@
...
+    """That the LatinHypercubeRandomPointSet initializes correctly"""
+    f = smolyay.samples.LatinHypercubeRandomPointSet(
+        [[-10, 20]], 64, 5678, True, "random-cd", 1
+    )
+    assert numpy.array_equal(f.domain, [[-10, 20]])
+    assert f.num_dimensions == 1
+    assert f.num_points == 64
+    assert f.seed == 5678
+    assert isinstance(f.scramble, bool)
+    assert f.scramble == True
```
mphoward: Same here, either this:
```suggestion
    assert f.scramble is True
```
or just `assert f.scramble`

mphoward (approved):
Looks great Michaela! I have a few small things to address, and then this can be merged.

smolyay/samples.py
mzbush: yes

cakieslich (approved):

PR #27, opened 2024-09-26
Feature/integral

mzbush:
Add property that adds the definite integral of the basis function over its domain.

### DIFF
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -181,10 +181,10 @@ class ChebyshevFirstKind(BasisFunction):
     @property
     def integral_over_domain(self):
         """float: basis function integrated over the domain."""
-        if self.degree == 1:
+        if self.degree % 2 == 1:
             return 0
         else:
-            return ((-1)**self.degree + 1)/(1-self.degree**2)
+            return 2/(1-self.degree**2)
 
     @degree.setter
     def degree(self, value):
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -26,10 +26,9 @@ class BasisFunction(abc.ABC):
         """numpy.ndarray: Domain the sample points come from."""
         pass
 
-    @property
     @abc.abstractmethod
-    def integral_over_domain(self):
-        """float: basis function integrated over the domain."""
+    def integrate(self):
+        """Integrate basis function over its domain."""
         pass
 
     def __call__(self, x):
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -647,6 +647,7 @@ class BasisFunctionSet(collections.abc.Sequence):
         result = numpy.array([x.integrate() for x in self.basis_functions])
         if domain is not None:
             result *= (domain[1]-domain[0])/(self.domain[1]-self.domain[0])
+        return result
         
 
 class ChebyshevFirstKindBasisFunctionSet(BasisFunctionSet):
--- a/smolyay/basis.py
+++ b/smolyay/basis.py
@@ -26,11 +26,6 @@ class BasisFunction(abc.ABC):
         """numpy.ndarray: Domain the sample points come from."""
         pass
 
-    @abc.abstractmethod
-    def integrate(self):
-        """Integrate basis function over its domain."""
-        pass
-
     def __call__(self, x):
         """Evaluate the basis function.
 
@@ -140,7 +135,11 @@ class BasisFunction(abc.ABC):
             Value of the derivative of the basis function.
         """
         pass
-
+    
+    @abc.abstractmethod
+    def integrate(self):
+        """Integrate basis function over its domain."""
+        pass
 
 class ChebyshevFirstKind(BasisFunction):
     r"""Chebyshev polynomial of the first kind.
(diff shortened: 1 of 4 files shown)


smolyay/basis.py
```diff
@@ -172,6 +178,14 @@ def degree(self):
         """int: Degree of polynomial."""
         return self._degree
 
+    @property
+    def integral_over_domain(self):
+        """float: basis function integrated over the domain."""
+        if self.degree == 1:
+            return 0
+        else:
+            return ((-1)**self.degree + 1)/(1-self.degree**2)
```
mphoward: The first case can cover all the odd degrees evaluating to zero, then the second case can be simplified because degree is even:
```suggestion
        if self.degree % 2 == 1:
            return 0
        else:
            return 2/(1-self.degree**2)
```

smolyay/basis.py
```diff
@@ -295,6 +309,11 @@ def domain(self):
         """numpy.ndarray: Domain the sample points come from."""
         return numpy.array([-1, 1])
 
+    @property
+    def integral_over_domain(self):
+        """float: basis function integrated over the domain."""
+        return (1 - (-1)**(self.degree+1))/(self.degree + 1)
```
mphoward: You can use odd / even to do this one too:
```suggestion
        if self.degree % 2 == 1:
            return 0
        else:
            return 2/(self.degree + 1)
```

tests/basis_test.py
```diff
@@ -345,6 +345,27 @@ def test_trig_initial():
...
+        (smolyay.basis.ChebyshevSecondKind(2), 2 / 3),
+        (smolyay.basis.Trigonometric(0), 2 * numpy.pi),
+        (smolyay.basis.Trigonometric(1), 0),
+        (smolyay.basis.Trigonometric(-1), 0),
+    ],
+    ids=basis_id,
+)
+def test_cheb_initial(basis_fun, integral):
+    """Test integral over the basis function's domain"""
+    assert basis_fun.integral_over_domain == integral
```
mphoward: These are technically floating point values, so it is recommended to use approx:
```suggestion
    assert basis_fun.integral_over_domain == pytest.approx(integral)
```

mphoward (changes requested):
Thanks! These are straightforward changes and can be merged after these suggestions are implemented.

smolyay/basis.py
```diff
@@ -26,6 +26,12 @@ def domain(self):
         """numpy.ndarray: Domain the sample points come from."""
         pass
 
+    @property
+    @abc.abstractmethod
+    def integral_over_domain(self):
+        """float: basis function integrated over the domain."""
+        pass
```
mphoward: I will put a suggestion below to call the method `integrate()` in the surrogates, and so I think the basis functions should also use that.
```suggestion
    @abc.abstractmethod
    def integrate(self):
        """Integrate basis function over its domain."""
        pass
```
It would make sense to move this method after the `derivative` method as well.

Revise as needed elsewhere based on this change.

smolyay/basis.py
```diff
@@ -533,7 +563,12 @@ def domain(self):
         if len(self._basis_functions) == 0:
             raise AttributeError("No basis functions to derive a domain.")
         return self._basis_functions[0].domain
-
+    
+    @property
+    def integral_over_domain(self):
```
mphoward: Rename and move
```suggestion
    def integrate(self):
```

smolyay/surrogate.py
```diff
@@ -631,6 +635,37 @@ def fit_gradient(self, X, y, X0=None, y0=None):
             self._integration_constant = integration_constant
         return self
 
+    def compute_definite_integral(self):
```
mphoward:
```suggestion
    def integrate(self):
```

smolyay/surrogate.py
```diff
@@ -631,6 +635,37 @@ def fit_gradient(self, X, y, X0=None, y0=None):
...
+        for ic, coeff in zip(self._index_combinations, self._coefficients):
+            answer += numpy.real(
+                coeff
+                * numpy.prod(
+                    [lookup_table[dim][ic[dim]] for dim in range(len(ic))], axis=0
+                )
+            )
+
+        # return results
+        return answer
```
mphoward: Maybe I am missing it, but I don't think I'm seeing a spot where the integral gets multiplied by a factor accounting for the domain of the surrogate being different than that of the basis functions? You should be able to take your answer as you've computed it and multiply it by this factor before you return it.

I also don't think I see a unit test covering this case.

mphoward (changes requested):
Thanks Michaela! Overall this looks pretty straightforward and good. Here are some comments that need to be addressed before merging.

smolyay/surrogate.py
mzbush: I hadn't added that feature yet. It's added now.

smolyay/basis.py
```diff
@@ -608,6 +632,23 @@ def derivative(self, X, domain=None, n=1):
...
+            the lower and upper bounds of X.
+            
+        Returns
+        -------
+        scalar or ndarray
+            the values of the basis functions."""
+        if domain is None:
+            return [x.integrate() for x in self.basis_functions]
+        else:
+            return [x.integrate()*(domain[1]-domain[0])/(self.domain[1]-self.domain[0]) for x in self.basis_functions]
```
mphoward: This method is supposed to return a NumPy array, but it doesn't currently. This would fit the pattern used above.
```suggestion
        result = numpy.array([x.integrate() for x in self.basis_functions])
        if domain is not None:
            result *= (domain[1]-domain[0])/(self.domain[1]-self.domain[0])
```

mphoward (approved):
Thanks Michaela! Please implement the small thing marked below, then this is good to go from my side.

mphoward (approved):

PR #28, opened 2024-09-26
Feature/newtoncotes

mzbush:
Add Newton-Cotes point set (for referencing)

### DIFF
--- a/smolyay/_growth.py
+++ b/smolyay/_growth.py
@@ -34,11 +34,9 @@ def make_clenshaw_curtis_level_sizes(num_levels):
     """
     rule = lambda x: 1 if x == 0 else 2**x + 1
     num_per_level = numpy.ones(num_levels, dtype=int)
-    num_per_level[1:] = [
-        rule(i) - rule(i - 1) for i in range(1, num_levels)
-    ]
+    num_per_level[1:] = [rule(i) - rule(i - 1) for i in range(1, num_levels)]
     return num_per_level
-    
+
 
 def make_slow_clenshaw_curtis_level_sizes(num_levels):
     r"""Nested levels for Clenshaw Curtis Slow Exponential Growth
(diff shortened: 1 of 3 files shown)


mphoward:
@mzbush I don't remember ever having looked at this before. What was the motivation for this PR, and do we think it actually needs to be merged into `refactor`?

mzbush:
@mphoward They were mentioned once in slack. To quote myself "The Newton-Cotes points are primarily for my own reference as these are the points that the regionally adaptive methods primarily use." When I was messing around with trying to replicate the regionally adaptive method from this paper here https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/wrcr.20467 I made some point sets for newton cotes since that's what they used so I could better understand it.

Adding equidistant, open newton cotes, and closed newton cotes were fairly trivial. (Equidistant in practice is just the nonnested Trigonometric but the inputs are different.)

Had the adaptive sampling development gone further, I felt it would be important to have newton cotes included since so many papers on adaptive methods use it and test if using an equidistant point set for adaptive grids is actually important when a nonequidistant point set could potentially be used instead. As such, i ended up opening a pull request to be dealt with in the far future to make sure I wouldn't lose that work.

Merging this with refactor would require me to finish writing the pytests for those new classes. It wouldn't take that long, but there's very little point in doing it right now. Should someone else pick up this project to work on adaptive sampling, they might see reason to finish this pull request so they can compare the adaptive results they get with other papers.

mphoward:
Ok, thanks! In that case, can you please make a copy of the relevant code for these sampling schemes and store it with the rest of your work related to the adaptive sampling project? Then we will plan to close this PR for now.

mphoward:
I will leave this branch around for a little bit if we decide it should be retargeted onto `main`.

PR #29, opened 2024-11-05
Feature/benchmark update

mzbush:
Makes the benchmark parent class input compatible with the Unidimensional and Multidimensional point sets


mphoward (approved):
Thanks Michaela! This looks good to me.

